PR #1, opened 2015-10-13
Minor tweaks

tianon:
The full diff here looks terrible, so I split this out into a series of much smaller commits, and I'm happy to elaborate on any of them.  The changes themselves were small enough conceptually (even if `diff` doesn't agree :innocent:) that I figured it'd be easier to discuss on a concrete PR instead of a series of back-and-forth comments. :smile:

This is the set of minor improvements I saw while reviewing for https://github.com/docker-library/official-images/pull/1070. :+1:

Overall, very pleased! :metal:

### DIFF
--- a/5.3/Dockerfile
+++ b/5.3/Dockerfile
@@ -2,29 +2,36 @@
 FROM    java:openjdk-8-jre
 MAINTAINER  Martijn Koster "mak-docker@greenhills.co.uk"
 
-ENV SOLR_VERSION 5.3.1
-ENV SOLR_SHA256 34ddcac071226acd6974a392af7671f687990aa1f9eb4b181d533ca6dca6f42d
-ENV SOLR solr-$SOLR_VERSION
+RUN apt-get update && \
+  apt-get -y install lsof && \
+  rm -rf /var/lib/apt/lists/*
+
 ENV SOLR_USER solr
 ENV SOLR_UID 8983
 
-RUN export DEBIAN_FRONTEND=noninteractive && \
-  apt-get update && \
-  apt-get -y install lsof && \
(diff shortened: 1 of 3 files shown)


makuk66:
Thanks for the PR! I'm happy to merge as-is, but some reactions:

- I had put the useradds at the end to ensure better caching of the solr download+check+install layer if a user wants to modify the Dockerfile to use a different userid locally, perhaps for volume mounting reasons. I assume you moved it to the top to get the benefit of caching if the solr version changes where the lsof does not. I'm fine with that.
- the idea behind the /opt/solr symlink was to closely follow typical solr practice: have a standard solr install which includes the version number, and then just have the symlink for local convenience so you can script `docker exec` things without having to know the version in use. The same applies to having the version variable path instead of just /opt/solr. But sure, losing the version number in the directory does make it look a bit simpler, so that's a good thing.
- the extra bash invocation -- oops I don't know what that was doing in there.
- The extra gpg verification in the Dockerfile itself -- I had considered that and prototyped it, but was not entirely keen on starting to receive keys from server and downloading asc files from urls in an automated Dockerfile build environment where I have no idea about what might be happening to that traffic. Doing it in the update.sh means it's in my environment and I can verify visually what's happening with the keys, and I check in the result where it can be tracked/inspected/verified in future. Ultimately, at docker build time, if someone manages to somehow forges the tarball, the asc file and the keyservers, it's the checked-in sha at will put a stop to it. I don't really see a scenario where the extra check can thwart an attack without the sha tripping it up first expect if the repo is forged, but by that point all bets are off because who knows what the forged Dockerfile could do. But, I guess it doesn't do any harm either, so sure, if you're happy with it, that's fine by me.

PR #16, opened 2016-04-01
Added alpine variants

LaurentGoderre:
This creates images half the size of the regular ones

/cc @sylus

### DIFF
--- a/5.3/alpine/Dockerfile
+++ b/5.3/alpine/Dockerfile
@@ -1,0 +1,39 @@
+
+FROM    java:openjdk-8-jre-alpine
+MAINTAINER  Martijn Koster "mak-docker@greenhills.co.uk"
+
+RUN apk add --update \
+        lsof \
+        gnupg \
+        tar \
+        bash \
+    && rm -rf /var/cache/apk/*
+
+ENV SOLR_USER solr
+ENV SOLR_UID 8983
+
+RUN addgroup -S -g $SOLR_UID $SOLR_USER && \
+  adduser -S -u $SOLR_UID -g $SOLR_USER $SOLR_USER
+
+ENV SOLR_KEY EDF961FF03E647F9CA8A9C2C758051CCA3A13A7F
+RUN gpg --keyserver ha.pool.sks-keyservers.net --recv-keys "$SOLR_KEY"
+
+ENV SOLR_VERSION 5.3.2
+ENV SOLR_SHA256 7c26601229ed712c639d00882f35304d87e0032028be4754d098a9b694877f48
+
+RUN mkdir -p /opt/solr && \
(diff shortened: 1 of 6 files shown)


makuk66:
Thanks. I like the idea, but I can't merge that as-is because:
- need a Dockerfile-alpine.template and modify update.sh so that alpine versions get created
- need to modify build-all.sh to build the alpine versions locally
- I expect generate-stackbrew-library.sh needs changing too
- The tar/wget changes -- did you change the argument ordering and short form just as a matter of taste?
- I expect update.md might need some changes, either in steps or in example output

LaurentGoderre:
Fair enough though this outlines how we would get there.

For the wget changes, alpine uses POSIX version of wget so the parameters are different. I modified the call to work with the builtin wget. For the tar, the POSIX was missing some of the option so I downloaded a different one than the default one.

LaurentGoderre:
@makuk66 I will work with you to get this working

makuk66:
Thanks for the info -- I'd like to keep the dockerfiles as same a possible, so we might just backport those changes to the non-apline versions too.

I expect we'll also need to install bash for some of the bin scripts; and we should verify it they still work.

It may keep the scripts cleaner if we keep the alpine versions on the same level rather than subdirectories.

I'll make some time for this on Monday.

LaurentGoderre:
I added bash to the install. 

For the hierarchy, I am doing what I saw done in the java and other images

LaurentGoderre:
Added the template and updated the update.sh

sylus:
This looks awesome to me! The only minor nit is I usually like seeing the packages each on a new line for readability. 

https://github.com/docker-library/php/blob/0d23b3d/5.6/fpm/alpine/Dockerfile#L4

I've also added @ncopa as he is doing all of the other official alpine images for docker-library in case has any points.

LaurentGoderre:
@makuk66 I wonder if this block will work as is for this case (changing the variant names of course)

https://github.com/nodejs/docker-node/blob/master/generate-stackbrew-library.sh#L33-L44

LaurentGoderre:
I regenerated all the dockerfile from the template and updated script. Only need to update the build-all and stackbrew

LaurentGoderre:
Updated the build-all script

makuk66:
The build-all.sh looks good, and works for me.
That generate-stackbrew-library snippet looks good, and works for me.
The update.sh looks good. Just stick a space after that comment #.
I've pushed a change to use the short-form `-O` for wget, to reduce the diff on this end.
The updated alpine tar can use the long form, so switch back to that.

LaurentGoderre:
I made the changes. I haven't done anything for the generate-stackbrew yet though

LaurentGoderre:
Also, some of the scripts don't have the executable mod in the commited files, is that intentional?

makuk66:
Thanks @LaurentGoderre; I've merged.

> Also, some of the scripts don't have the executable mod in the commited files, is that intentional?

No; fixed.

Issue #14, opened 2016-02-29
How to create_core in Dockerfile or with docker-compose?


oppianmatt:
I don't want to have to create_core each time, how do run it as part of the Dockerfile? Can I COPY/ADD some config and have a core ready to go?

oppianmatt:
seems also that solr needs to running to create a core?

My issue is that I want to create a dockerized solr in CI without having to manually create_core.

oppianmatt:
The docs say to create a core you run docker exec which attaches to an existing container instead of docker run which would start a command in a new container. That is needed because solr isn't exporting any volumes. If solr exported the volumes docker run would work.

Issue is that there is no docker exec for docker-compose. There is a docker-compose run but it starts a new container.

The way is to get docker-compose to give the list of service and run docker exec against that:

```
docker exec -it $(docker-compose ps | grep <service> | cut -d" " -f 1) <command>
```

But also people are thinking that create_core is a stand alone command but it's not. It requires solr to be running because it creates cores by making a making a http request like so:


```
http://localhost:8983/solr/admin/cores?action=CREATE&name=foo1&instanceDir=foo1
```

But for those who want to know how to create a core without having a running instance, extend the Dockerfile:

```
FROM makuk66/docker-solr
WORKDIR /opt/solr/server/solr/
ADD . /opt/solr/server/solr/
```

Then you need any directory with a core.properties file (can be empty). And from there have a conf directory with schema.xml and solrconfig.xml. Although with the new autoschema you might not need it.

oppianmatt:
One more small issue is that ADD does not respect USER so you need to chown the files like so:

```
FROM makuk66/docker-solr
WORKDIR /opt/solr/server/solr/
ADD . /opt/solr/server/solr/
USER root
RUN chown -R solr /opt/solr/server/solr
USER solr
```

makuk66:
> I don't want to have to create_core each time, how do run it as part of the Dockerfile? Can I COPY/ADD some config and have a core ready to go?

I'd do it like this: First run up a temporary Solr container and create the core:
```
docker run -d -P --name solr1 solr
docker exec -it --user=solr my_solr bin/solr create_core -c gettingstarted
```
then stop Solr and export the created core directory:
```
docker exec -it --user=solr solr1 bin/solr stop
docker cp solr1:/opt/solr/server/solr/gettingstarted solr1_gettingstarted
```
Then use a custom Dockerfile to build your image:
```
cat > Dockerfile <<EOM
[...]

PR #26, opened 2016-04-27
Foreground init

makuk66:


tianon:
This is better, but still suffers from the problem that all the `initdb` scripts have no way other than a background process that waits of actually connecting to the running database reliably (and if one of them fails, the process doesn't fail).  I'm also not sure I completely understand the use case for the `solr-demo`, etc. shortcuts -- can you elaborate on how those are expected to be used a little bit?  (Since it seems to me that it'd be a more straightforward experience for users to either provide a simple `initdb` script to perform some of those actions or to start up the main database and perform those from a separate container.)

makuk66:
> all the initdb scripts have no way other than a background process that waits of actually connecting to the running database reliably (and if one of them fails, the process doesn't fail). 

An initdb script can make changes to the disk configuration before Solr gets started as normal.
An initdb script can start Solr in the background, talk to it from the foreground, do whatever error-checking and reporting it wants (and exit if needed), and then stop it, ready to get started again in the foreground as normal. Both satisfy your criteria as I understand them. I longer have any documentation or examples of scripts that run in the background while Solr runs in the foreground. What am I missing?

> I'm also not sure I completely understand the use case for the solr-demo, etc. shortcuts -- can you elaborate on how those are expected to be used a little bit? (Since it seems to me that it'd be a more straightforward experience for users to either provide a simple initdb script to perform some of those actions or to start up the main database and perform those from a separate container.)

See https://github.com/docker-solr/docker-solr/issues/14. To do anything interesting with Solr, you need to create a index. You cannot create a Solr core solely through the UI or API at this time; some on-disk work is required. So it can't be done from another container. Solr provides a command-line utility to do this, but that has to run on the same container, so currently requires `docker exec` (which is what we used to recommend), or writing external scripts mapped into the container (as a volume) or from a custom Dockerfile, or writing long command-lines. That is inconvenient and error-prone. Using the `solr-create` and `solr-precreate` commands in the docker entrypoint you can do it simply in a single `docker run`, and have a container ready to use from the UI or remotely through the API from some other container.

The `solr-demo` is simply an extension of the above that loads some of Solr's provided example data into a core, so you can start showing off Solr searching with a single command. You could do that remotely from a second container (if you start the first one with a `solr-create` or `solr-precreate`), or you could do it from an initdb script. But surely that is less straightforward.

scripts/wait-for-solr.sh
```diff
@@ -1,8 +1,7 @@
 #!/bin/bash
 #
 # A helper script to wait for solr
-set -x
-until $(wget -O - http://localhost:8983 | grep -q -i solr); do
+until $(wget -q -O - http://localhost:8983 | grep -q -i solr); do
```
tianon: This is currently taking the output of `wget | grep -q` and trying to run it as a command, so can be replaced with simply:
```bash
until wget -q -O - http://localhost:8983 | grep -q -i solr; do
```

tianon:
Ok, fair enough -- thanks for taking the time to explain. :+1:

LGTM (although one minor comment on the `wait-for-solr.sh` script)

Issue #28, opened 2016-05-12
wait-for-solr should give up after a while


makuk66:
Per https://github.com/docker-library/official-images/pull/1637#issuecomment-218619452 

```
you may want a configurable number of retries on the wait-for-solr script so the the container will kill itself if solr is misconfigured and never actually comes up.
```

PR #30, opened 2016-05-16
wait-for-solr.sh now gives up after 1 minute

makuk66:
Fixes https://github.com/docker-solr/docker-solr/issues/28


yosifkit:
:+1: I think this looks like a solid plan.

makuk66:
Thanks for the review!

PR #37, opened 2016-06-08
fix typo in arguments, this should be volume

saily:
This looks like a typo to me, i assume this should be a volume mount


makuk66:
Thanks!

PR #54, opened 2016-09-05
Remove docs/ (82.5MB)

dsmiley:
Lets keep the images lean, please.

Maybe there's something to be said about removing dist/* & contrib/* as well... perhaps only on the alpine image.  Someone who chooses Alpine wants lean-ness.


makuk66:
I agree with removing the docs, as they are not exposed through the UI anyway (the "Documentation" link in the footer refers to http://lucene.apache.org/solr/).

I'm less convinced about the dist and contrib, as they are referenced from the configuration.
I think there's an argument to made that they shouldn't be, but then I feel that'd be more a matter for the upstream packaging.

makuk66:
The build failure here is unrelated to the content of the PR. This problem is tracked https://github.com/docker-solr/docker-solr/issues/56

Issue #55, opened 2016-09-05
Options should be settable from environment variables


dsmiley:
I've noticed other docker-ized apps that use environment variables to toggle settings.  I think this is _much_ more user friendly than asking folks to modify `solr.in.sh` in the container.  Despite docker's awesomeness, modifying state is it's Achilles' heal (IMO).  Fortunately, Solr's bin script is already designed around a lot of useful environment variables.  We just need to remove or rename (so it can be referred to still) `solr.in.sh`.   That mechanism would be more useful generally if it was written to check that each var is undefined before defining it.  In such a way, this would make Solr easier to use outside of docker as well since I could simply set some env vars to configure Solr if I chose to do it that way -- knowing solr.in.sh wouldn't stomp on them. Others have asked about this too: #53 #44

makuk66:
> I've noticed other docker-ized apps that use environment variables to toggle settings. I think this is much more user friendly than asking folks to modify solr.in.sh in the container. Despite docker's awesomeness, modifying state is it's Achilles' heal (IMO).

Agreed, this would be a good thing to have.

> Fortunately, Solr's bin script is already designed around a lot of useful environment variables. We just need to remove or rename (so it can be referred to still) solr.in.sh. That mechanism would be more useful generally if it was written to check that each var is undefined before defining it. In such a way, this would make Solr easier to use outside of docker as well since I could simply set some env vars to configure Solr if I chose to do it that way -- knowing solr.in.sh wouldn't stomp on them.

Also agreed, and that has bitten me outside Docker in the past too.
But that sounds best addressed in the upstream packaging.

Also relevant: [SOLR-7850](https://issues.apache.org/jira/browse/SOLR-7850)


I believe that the docker-solr image should in general be as much like upstream as possible, to remove needless confusion when users switch to docker. So, if we want to implement a better environment configuration mechanism, its default configuration should be the same. Currently, solr.in.sh defines various environment variables that we would need to replicate, and that increases the maintenance headache when the upstream contents  change. Instead or replicating we could I suppose to do some environment variable renaming (rename solr.in.sh to solr.in.sh.dist, write our own solr.in.sh, in it save any overridden variables under a different name, then load the  solr.in.sh/dist, then apply the overrides); that way we inherit value changes are least, but could still miss out on new values.

makuk66:
How about something like that:

```
[...]

PR #60, opened 2016-09-07
Issue55 env settings

makuk66:
See https://github.com/docker-solr/docker-solr/issues/55


makuk66:
With Solr 6.3 this is no longer required.

PR #66, opened 2016-09-20
Change default CONFIG_SOURCE to upstream default

IlyaSemenov:
docker-solr uses `basic_configs` as the default config source (see https://github.com/docker-solr/docker-solr/blob/master/6.2/scripts/docker-entrypoint.sh#L96), however the upstream default is `data_driven_schema_configs` (see https://github.com/apache/lucene-solr/blob/branch_6_2/solr/bin/solr#L751 - it has been like this at least since Solr 5.4).

As the result, Solr cores created with `docker run ... solr-precreate <core>` behave differently from those created locally with `solr create -c <core>`.

This PR fixes the inconsistency.


makuk66:
I'm all for following upstream, and for consistency.
I think we'll want to:
- call this out in the upstream PR, as it is a potentially breaking change
- add a FAQ explaining you can get the old behaviour by providing the directory to the command
@shalinmangar does this look OK to you?

shalinmangar:
+1 LGTM

PR #69, opened 2016-09-30
Update README to reflect Apache branding

dsmiley:
I think the first heading with "Solr" in it should clearly show as "Apache Solr".  After the heading and first sentence, the remaining reference needn't.  I'm not certain if the page was a problem as it was (in the eyes of the ASF) but it should be a-okay now.  In addition, I added a reference to the ASF governance... this is after all a key differentiator between Solr and it's competitors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,9 +13,9 @@ For more information about this image and its history, please see [the relevant 
 
 For detailed information about the virtual/transfer sizes and individual layers of each of the above supported tags, please see [the `solr/tag-details.md` file](https://github.com/docker-library/docs/blob/master/solr/tag-details.md) in [the `docker-library/docs` GitHub repo](https://github.com/docker-library/docs).
 
-# What is Apache Solr?
+# What is Apache Solr™?
 
-Apache Solr™ is highly reliable, scalable and fault tolerant, providing distributed indexing, replication and load-balanced querying, automated failover and recovery, centralized configuration and more. Solr powers the search and navigation features of many of the world's largest internet sites.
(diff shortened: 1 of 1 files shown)


makuk66:
Per http://www.apache.org/foundation/marks/pmcs.html#attributions the first mention seems to want a ™ as well.

The introductory paragraph came verbatim from http://lucene.apache.org/solr/. I don't see the need to call out the ASF in an introductory paragraph -- if "Apache" is in the project name it's clearly an Apache project, and the ASF is implicit in that. And for those who don't know what the ASF is, just mentioning it doesn't help much without a link. On the http://lucene.apache.org/solr/ the ASF is not mentioned until right at the end, but does at least do it justice. I'm happy to leave explanations of Solr to its home page.

dsmiley:
+1 to the TM.

Are you okay with the heading?

RE the first paragraph; yeah I can see why you simply took it verbatim -- why describe Solr yourself when it already does so :-)  I was just trying to improve it a little.  I'm not sure if _everyone_ knows precisely what Apache is.

makuk66:
There are lots of other great aspects of Solr too, which equally could be argued to be important enough to warrant the intro (it offers much beyond simple querying, it's Java, it has an API, it's robust at 12 years old, it has a great community etc etc); but we need to balance that against becoming a wall of opinionated text. FWIW, I'm not a fan of that whole paragraph, and I'm sure you could debate such things endlessly; but if that's what the project uses, then that's what I want to stick to. After all, this README is about the packaging for Docker, and should focus on that.

I've added the TM and removed that line. I'm happy to merge as-is.

dsmiley:
+1

Issue #71, opened 2016-10-06
Can't use "solr zk" ?


dsmiley:
I'd like to use the zk upload capability.  I know this isn't the complete invocation but if I run this:

```
$ docker run --rm solr zk
```

I get:
/opt/docker-solr/scripts/docker-entrypoint.sh: line 131: exec: zk: not found

And if I try:

```
$ docker run --rm solr solr zk
```

I get:
ERROR: zk is not supported by this script

Usage: solr start ....etc

makuk66:
Yeah, that's not ideal. We'll have to think about how best to fix that.

In the meantime, as a workaround you can do:

```
docker run -it solr bin/solr zk
```

makuk66:
@dsmiley Have a look at that PR; I think this is probably as good as we can make it: if you want to run solr as a server as intended, you can omit the command name. If you want to run solr for some other purpose, you'll have to specify the "solr" command explicitly (so you end up with "solr solr") and provide whatever arguments you desire.

This is a possibly breaking change: for users who currently run with "docker run solr solr" and rely on the `-f` being provided will be caught out. I'm not too happy about that.

I'm also not entirely wild about this use of the solr command. A lot of the help text and documented use cases make assumptions about this command operating on the local Solr -- which there isn't here because you're in a container that is separate from any other. Most commands won't make any sense in this context (eg stop, restart, status) and others that do (eg zk healthcheck) will need special arguments to connect to the right zookeeper/solr server. So is this actually useful?

dsmiley:
FYI I'm at LSR conf and will look closer when I get back; perhaps over the weekend.  But in essence... my use-case is with mounting a volume to where my config is on disk.  Users should be able to do this stuff purely with Docker without having to download a Solr distribution the non-Docker way.

Once I figured out I could do `bin/solr` I was happy... it was just non-obvious and it took some poking around to figure this out.

dsmiley:
The PR LGTM.

dsmiley:
BTW, I have no convictions on the best way to solve this issue but I thought I'd mention that I think a fine approach is simply to return a helpful error message if a known command is invoked in an unsupported way.  That may be the easiest and perhaps involves the least "magic"; not sure.

PR #73, opened 2016-10-13
Make "solr zk" work

makuk66:
Change the default CMD to "solr-foreground", and run solr with `-f` only for that.

For example:

To run an arbitrary command:

```
docker run -it solr /bin/date
```

To run solr in the foreground without arguments:

```
docker run -it solr
```

To run solr in the foreground with extra arguments:

```
docker run -it solr -h myhostname
```

To run solr as an arbitrary command:

```
docker run -it solr solr zk --help
```

Fixes https://github.com/docker-solr/docker-solr/issues/71

### DIFF
--- a/5.3/Dockerfile
+++ b/5.3/Dockerfile
@@ -53,4 +53,4 @@ WORKDIR /opt/solr
 USER $SOLR_USER
 
 ENTRYPOINT ["docker-entrypoint.sh"]
-CMD ["solr"]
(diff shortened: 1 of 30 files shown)


dsmiley:
Nice.  I didn't try it myself but if it works as you listed then LGTM.

makuk66:
closed in favour of https://github.com/docker-solr/docker-solr/pull/83

Issue #74, opened 2016-10-17
initialize solr.xml if empty SOLR_HOME dir


dsmiley:
When using Solr in Docker in production (or anywhere I guess), I think it's common to set the SOLR_HOME environment variable to some volume.  On first use, that volume will be empty, and Solr complains about a missing solr.xml.  It can be annoying to put a template one into place there a-priori.    In a project I've worked on, I customized my docker file to execute this script:
https://github.com/dsmiley/hhypermap-bop/blob/master/solr-docker/initEmptySolrHome.sh
It checks if it's not there and if puts it there.

BTW thanks for including the feature in docker-solr to execute shell scripts placed in `/docker-entrypoint-initdb.d/ -- that's very useful!`

I should reference https://issues.apache.org/jira/browse/SOLR-9575   My shell script could be improved to check that the directory is completely empty.

makuk66:
In docker-solr there is a `/opt/solr/server/solr/mycores` directory -- if you mount your volume there, then you can use the "solr-precreate" (see https://github.com/docker-solr/docker-solr/blob/master/README.md#creating-cores) to copy needed configset.
Does that achieve what you need here?

If some empty-dir-initialisation is to be done, I'd prefer for it to happen in Solr itself rather than docker-solr doing something non-standard. But I share Hoss's and Erick's concerns raised in SOLR-9575 about doing things based on SOLR_HOME which can lead to mistakes and hard to diagnose problems.

I'm also not wholly convinced that the "copy the whole solr dir contents" is right for production -- you probably don't need the sample_techproducts_configs, and you may not need the zoo.cfg, and you might want to tweak the settings in solr.xml; Using the `docker-entrypoint-initdb.d` you can create exactly what you do need.

dsmiley:
> In docker-solr there is a /opt/solr/server/solr/mycores directory -- if you mount your volume there, then you can use the "solr-precreate" (see https://github.com/docker-solr/docker-solr/blob/master/README.md#creating-cores) to copy needed configset.
Does that achieve what you need here?

[...]

PR #77, opened 2016-10-22
Issue74 solr home2

makuk66:
Possible support for intialising an empty SOLR_HOME; should we want to do that.

Candidate for https://github.com/docker-solr/docker-solr/issues/74


makuk66:
> BTW it looks like a pain to have copies for older Solr versions. Why maintain them at all like this? There are git tags, and branches could be used to do older releases.

Because that's how the docker-library/official-images project automated build consumes them.

PR #81, opened 2016-11-09
Dockerfile: add VOLUME for solr home dir

dsmiley:
It's useful metadata to declare the volume.  Other stateful services do this (e.g. Zookeeper docker images).  

Until recently I've yet to figure out what effect setting the VOLUME has without expressly mapping it.  Today I found one such effect in my use of Kontena.  It is unable to automatically create a data container for the service (declared to be stateful) because there are no volume paths to put in it.

### DIFF
--- a/Dockerfile-alpine.template
+++ b/Dockerfile-alpine.template
@@ -56,6 +56,8 @@ RUN chown -R $SOLR_USER:$SOLR_USER /opt/docker-solr
 
 ENV PATH /opt/solr/bin:/opt/docker-solr/scripts:$PATH
 
+VOLUME /opt/solr/server/solr/
+
 EXPOSE 8983
 WORKDIR /opt/solr
 USER $SOLR_USER
--- a/Dockerfile.template
+++ b/Dockerfile.template
@@ -48,6 +48,8 @@ RUN chown -R $SOLR_USER:$SOLR_USER /opt/docker-solr
 
 ENV PATH /opt/solr/bin:/opt/docker-solr/scripts:$PATH
 
+VOLUME /opt/solr/server/solr/
+
 EXPOSE 8983
 WORKDIR /opt/solr
 USER $SOLR_USER


makuk66:
https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#volume agrees with you. I wonder if this would cause any changed behaviour in other scenarios.

makuk66:
I've looked into it some. Using VOLUME in a Dockerfile is not just metadata; it changes behaviour. Per https://docs.docker.com/engine/reference/builder/#volume, it creates an anonymous volume at `docker run` time, and copies the existing contents of that location into that volume. In our case that's 1.5M of Solr config. When your container exits, and you `docker rm` the container, that volume does not get destroyed (unless you pass an explicit `-v` to `docker rm`, or you specified `--rm` to the `docker run`) and you then need to manually cleanup these dangling volumes (see `docker volume ls -qf dangling=true`). This is especially jarring when you run a container for a reason other than running a server (say bash), where the extra volume copy pointless. These volumes can't be be consumed explicitly other than by using `docker run --volumes-from` -- but if you want to share volumes, the newer "docker volume create --name mynamedvol" named volumes are clearer and easier to use (although admittedly you do have to then do the copy yourself). If you really want the aforementioned behaviour, you can get it at run-time by specifying `docker run -v /opt/solr/server/solr/` so you have the choice, whereas if you use the VOLUME instead there is no way to disable that at run-time.
So while I can see why VOLUME used to be advisable, I'm not convinced that is still the case.

I'm not familiar with Kontena. Can you achieve your use-case by using named volumes (perhaps fro Compose), or by specifying a `docker run -v /opt/solr/server/solr/`. If not, can you explain more about how this works in Kontena please?

dsmiley:
Excellent investigation -- thanks for shedding light on this!

Given that Solr ships with it's config sets underneath SOLR_HOME, and they weigh a non-trivial size, I think we shouldn't declare the VOLUME after all.  As you indicated, it's easy enough to specify the volume outside the Dockerfile.  Perhaps these should be added nonetheless as commented with a comment as to _why_ so that others understand.

Longer term, perhaps Solr upstream could change the location of it's built-in config sets; perhaps to be at /opt/solr/configsets or opt/solr/server/configsets (debatable).  I suspect there will be resistance.

RE Kontena -- indeed I can specify the named volume in docker-compose.yml which Kontena understands.  I've already filed an enhancement issue with Kontena for it to complain when declaring a service as stateful (a Kontena feature) when no volumes are declared anywhere (no volumes in docker-compose or Dockerfile).

makuk66:
> Perhaps these should be added nonetheless as commented with a comment as to why so that others understand.

There are many things we don't do, documenting them all would be a bit much :-) For this particular issue, you can't condense all the implications into a single line description, and there are various alternative approaches one could consider. I'd rather keep the Dockerfile as concise as it can be; this is more something for a FAQ entry where there is more room (and can be modified without implications for the upstream processing). I think the existing volume-related FAQs need to be reviewed (recommend `docker volume` rather than the data volume containers, and refer to the README's compose and SOLR_HOME examples), perhaps it could be worked into that.

dsmiley:
+1 to FAQ & Docker-compose example.

RE comment:
I needn't be comprehensive, just a simple statement as follows:

    # Commented volume since Solr's configsets are a bit large to do this by default. See #81
    # VOLUME /opt/solr/server/solr/

PR #83, opened 2016-11-11
Make "solr zk" work

makuk66:
Change the default CMD to "solr-foreground", and run solr with '-f' only for that.

For example:

To run an arbitrary command:

    docker run -it solr /bin/date

To run solr in the foreground without arguments:

    docker run -it solr

To run solr in the foreground with extra arguments:

    docker run -it solr -h myhostname

To run solr as an arbitrary command:

    docker run -it solr solr zk --help

Fixes https://github.com/docker-solr/docker-solr/issues/71

### DIFF
--- a/Docker-FAQ.md
+++ b/Docker-FAQ.md
@@ -18,11 +18,11 @@ For some use-cases it is convenient to provide a modified `solr.in.sh` file to S
 For example to point Solr to a ZooKeeper host:
 
 ```
-docker create --name my-solr -P solr
-docker cp my-solr:/opt/solr/bin/solr.in.sh .
+docker create --name my_solr -P solr
+docker cp my_solr:/opt/solr/bin/solr.in.sh .
 sed -i -e 's/#ZK_HOST=.*/ZK_HOST=cylon.lan:2181/' solr.in.sh
-docker cp solr.in.sh my-solr:/opt/solr/bin/solr.in.sh
-docker start my-solr
+docker cp solr.in.sh my_solr:/opt/solr/bin/solr.in.sh
+docker start my_solr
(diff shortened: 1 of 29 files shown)


makuk66:
@tianon I'm changing the entrypoint handing so that we can distinguish between running solr as a server, and as a generic command. Could you give this a once-over please?

Docker-FAQ.md
```diff
@@ -18,11 +18,11 @@ For some use-cases it is convenient to provide a modified `solr.in.sh` file to S
 For example to point Solr to a ZooKeeper host:
 
 ```
-docker create --name my-solr -P solr
-docker cp my-solr:/opt/solr/bin/solr.in.sh .
+docker create --name my_solr -P solr
+docker cp my_solr:/opt/solr/bin/solr.in.sh .
 sed -i -e 's/#ZK_HOST=.*/ZK_HOST=cylon.lan:2181/' solr.in.sh
```
dsmiley: Do we really want to make it look this hard vs. simply setting `-e ZK_HOST=CYLON.LAN:2181` ?

makuk66: For the purposes of this particular patch, this particular change changes the container name to use underscore instead of a dash. I think I actually ran into some reason that the dashed container name caused me a problem, but I don't recall what that was now. Perhaps I just changed this for consistency with Docker's own namesgenerator scheme, and the other examples in the README. In any case, that is fine.

That FAQ question is specifically about persisting config, and the answer illustrates how to write config to the container. If we use `-e` then there is no persisting, and the answer thus becomes non-sensical. Perhaps we can come up with a better use-case.

dsmiley: Oh I see.  I agree, find a better use-case that requires this approach so we don't make it seem hard to do common things.

PR #85, opened 2016-11-15
doc improvements

makuk66:


README.md
```diff
@@ -62,21 +62,55 @@ $ docker run --name solr_demo -d -P solr solr-demo
...
+$ docker exec -it --user=solr my_solr post -c gettingstarted mydata/mydata.xml
 ```
 
-or by mounting a host directory as a volume:
+which has the disadvantage that the host directory is mounted during the entire lifetime
+of the Solr container.
+
+If you use a remote Docker host, and thus cannot use mounted host directories, you can
+use a data volume instead. Here I use a temporary container for the file copy so that
+I can still mount the data volume read-only in Solr.
```
dsmiley: I'm missing the point of this example.  Are you assuming a remote docker-machine wherein your Solr's bin/solr post must also be remote and thus won't have access to local data?  Why not run bin/solr post from a local docker?

makuk66:
Sent from my iPhone

> On 15 Nov 2016, at 18:44, David Smiley <notifications@github.com> wrote:
> 
> @dsmiley commented on this pull request.
> 
> In README.md:
> 
> >  ```
>  
> -or by mounting a host directory as a volume:
> +which has the disadvantage that the host directory is mounted during the entire lifetime
> +of the Solr container.
> +
> +If you use a remote Docker host, and thus cannot use mounted host directories, you can
> +use a data volume instead. Here I use a temporary container for the file copy so that
> +I can still mount the data volume read-only in Solr.
> I'm missing the point of this example. Are you assuming a remote docker-machine wherein your Solr's bin/solr post must also be remote and thus won't have access to local data?
> 
Right
> Why not run bin/solr post from a local docker?
> 
You may not have a local docker daemon. For months, before there were native Docker daemon implementations on OSX/Windows, I would run a local docker client on OSX on my poor laptop, against a snazzy Linux server. That scenario is well suited to docker data volumes: you copy it once, it's then in the Docker world, and then can use it repeatedly against different containers. This approach might make sense in a production system too, especially if you use a volume provider that can be mount on multiple containers across hosts, and run concurrent posts on each.

If you _can_ mount a host directory locally, then that may be more appealing, and the earlier examples illustrate that. You would need make sure you adjust your docker commands to point at the right daemon (remote for Solr and getting the address, then local for the post)

makuk66:
Actually I want to do this differently altogether.

PR #93, opened 2017-01-04
Update set-heap.sh

afaulconbridge:
Because the SOLR_HEAP= line is commented out in the default /opt/solr/bin/solr.in.sh file the example set-heap.sh doesn't actually apply it to Solr.

On this pull request note the # added before the first SOLR_HEAP on the sed command that uncomments that line and therefore makes it used by solr

### DIFF
--- a/docs/set-heap.sh
+++ b/docs/set-heap.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
 set -e
 cp /opt/solr/bin/solr.in.sh /opt/solr/bin/solr.in.sh.orig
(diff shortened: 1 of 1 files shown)


dsmiley:
This is obsolete.  Since Solr 6.3, the SOLR_HEAP (and perhaps all) config should be set via environment variables.  I helped change Solr as-such to make it more Docker-friendly.

afaulconbridge:
Ah, that is indeed more Docker-friendly! It seems that the documentation needs to be updated then.

Issue #94, opened 2017-01-10
Init Actions have no running Solr instance


socketwench:
I was hoping to use init actions to augment the container to automatically create a core on first startup. This would be similar to the official mariadb container's init scripts. Unlike the mariadb container, however, the this container does *not* have a running instance of Solr at the time the init scripts are executed.

# Proposed solutions

1) Wrap the script execution in init_actions() with initial_solr_begin() and initial_solr_end(). 
Or
2) Create a "start actions" phase so as not to break existing init action scripts used in other containers. Start actions would work the same way as init actions, but would execute with a running Solr instance in the background.

I've already implemented Option 2 here: https://github.com/ten7/dockstack/tree/master/.docker/solr/5.5

makuk66:
> create a core on first startup

Looking at https://github.com/ten7/dockstack/blob/master/.docker/solr/5.5/100-create-core-from-env.sh I see that your use-case is that you want to create a collection based on a config you install at run-time. I think you can achieve that with the existing code already: Do the `wget`/`tar` part with `docker-entrypoint-initdb.d` to put the config in place in the container, while providing the normal https://github.com/docker-solr/docker-solr#creating-cores arguments to `create`/`pre-create` referencing said config. Am I missing something that would make that not work?

I think I see what you are trying to do in https://github.com/docker-solr/docker-solr/pull/95, but:

- I'm not yet convinced that it is needed, per above
- You changed solr-foreground to do the extra run-solr-in-the-background-then-quit, thus imposing that cost and log confusion onto everyone regardless of whether they need it. At the very least you should skip all that unless there are files in `docker-entrypoint-start.d`. But I think solr-foreground should do what its name says, and not do the double start.
- The docker-entrypoint-start.d mechanism only works with the solr-in-the-background-then-quit (ie currently `solr-create`), not the other `pre-create`/`solr-demo` flows, and that makes it inconsistent and thus possibly more confusing
- This adds yet more code into `docker-entrypoint.sh`, and invents yet another extension mechanism to have to understand. If anything I'd like to simplify.

I've been thinking about refactoring `docker-entrypoint.sh` such that you can more easily extend using composition, rather than modification and hooks. It could look something like

```
docker run -v $(PWD)/docker-entrypoint-start.d:/docker-entrypoint-start.d solr \
  bash -c "solr-local-start; run-parts /docker-entrypoint-start.d/; solr-local-stop; solr-foreground"
```
or better yet just create a `run.sh` that has:

```
[...]

PR #95, opened 2017-01-10
Enable customizable "Start actions" with a running Solr instance in the background. 

socketwench:
See https://github.com/docker-solr/docker-solr/issues/94


makuk66:
See the comments in the associated issue.

PR #96, opened 2017-01-14
split docker-entrypoint.sh into scripts

makuk66:
Split docker-entrypoint.sh into scripts so that it is easier to create your own scripts by combining pieces. See for example how `solr-demo` uses `run-initdb`, `start-local-solr`/`stop-local-solr`. This also considerably simplifies the `docker-entrypoint.sh`.


tianon:
Definitely like the idea here (taking your previous "fake" commands and making them legit commands that the user can invoke and combine in their own ways).

Would it be safe to trust that the content/function of the subcommands is the same, or were you looking for me to do more than an overall idea review here?


(hope you don't mind me replying here rather than in the email thread)

> our PR diffs are hard to read because we have copies of dockerfile/dockerentrypoint/scripts for each version and variant. I have it this way because I know that's how the library will build it. Is there a smarter way to do this, perhaps using a script that can populate directories at build time?

Not at build time from the official images context, but one approach could be to have your CI (Travis) run `./update.sh` before testing the images, and then your PRs only touch the template copies.  You could then setup either Travis or some other automated process to make commits to `master` post-merge which update all the copies.

At the risk of probably sounding like a broken record (:sweat_smile:), it's probably also worth surveying again whether all the versions supported here are still actively supported by the Solr project at large (and receiving at least security updates, if not bugfixes as well).

makuk66:
Thanks for the review!

The content/function of the subcommands should be the same. More eyes on it can never hurt of course.

Having CI do the update is an intriguing idea; I'll have to experiment with how that best fits into my local workflow.

Point noted about the versions -- I'll look into that again.

PR #98, opened 2017-01-25
Fix positional arguments for solr-precreate script

tobiashm:
After being extracted from the docker-entrypoint.sh script, the positional arguments should have been shifted once.


tobiashm:
You currently have to add a dummy argument, like:
```yaml
version: '2'

services:
  solr:
    image: solr
    ports:
      - 8983:8983
    entrypoint:
      - docker-entrypoint.sh
      - solr-precreate
      - ""
      - mycore
```

makuk66:
Thanks for the fix!

Issue #106, opened 2017-02-20
Build fails due to missing gpg key


makuk66:
I tried to update for the latest 5.5.4 release, but that failed because the release engineer's code signing key is not on the public key servers. This is not the first time that has happened, and the release process https://wiki.apache.org/solr/HowToReleaseSlowly#Making_a_release does not prescribe adding your key to the public servers either, only to the Apache keys file, so it's no wonder.
And occasionally the key servers are unavailable, which can be hard to debug because we use a pool, so don't always talk to the same server.
Every time there is an issue like this it is a time sink.

So we should probably stop relying on the public key servers, and just rely on the Apache keys file instead. Or possibly fetch the keys file at `update.sh` time and then check it into the repo, and at build time consume it from there. We could also consider removing the GPG verification at build time and just rely on the checksum instead, thus doing GPG verification at `update.sh` time only, not at build time.

dsmiley:
I confess I don't understand why this part of the Dockerfile, the gpg signing, is even there.  Isn't the sha1 verification good enough?

makuk66:
The reason it is in there is because that was recommended as "best" by https://github.com/docker-library/official-images/blob/master/README.md#image-build
From my perspective a couple of things appealed:

- Because the gpg verification is in the Dockerfile it happens at the docker hub build servers -- so even if my workstation/github/CI infrastructure gets compromised, it can be detected there. But I struggle to think of a scenario where such compromise cannot also circumvent the verification.
- If you use gpg instead of the checksum, it might be easier for folks with cloned repos to update without having to run update.sh. But that's a niche use-case, and we do use a checksum so doesn't work.

Personally I think the checksum is sufficient.

makuk66:
I've committed the minimal change for now. If we need to revisit this again, we can

tianon:
Please do revisit this -- the signature and the checksum verify different things.  The signature verifies that the thing we've fetched is indeed published by the folks we expect it to be, where the checksum merely verifies that the bits haven't been tampered with since the checksum was embedded into the `Dockerfile`.

The main reason we prefer to use the keyservers rather than simply downloading the `KEYS` file is that it then also verifies the key itself is what we expect (which we get by using the full fingerprint as has been used in this repository previously); GPG, after fetching the key, will then verify that the key it received actually matches the full fingerprint provided (and we don't get that additional verification with this method -- if someone compromises the Apache distribution server, they can add their own key into the `KEYS` file at the same time as modifying the release tarball).

[...]

PR #113, opened 2017-03-08
Switch back to fetching keys from a keyserver, but use multiple keyservers to handle edge cases better

tianon:
Also, this modifies `update.sh` to ensure that several keyservers definitely have a copy of the public key we need.

This should account for flakiness in ha.pool.sks-keyservers.net, as well as flakiness in environments where the default HKP port is blocked (by explicitly trying to use other keyservers that listen on port 80).

This reverts some of https://github.com/docker-solr/docker-solr/pull/109 (be05c8c5748e588d38cb0ed96770cf7ba9956589), in reference to https://github.com/docker-solr/docker-solr/issues/106#issuecomment-285137497.

cc @yosifkit

ref https://github.com/docker-library/official-images/pull/2729

### DIFF
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -16,18 +16,27 @@ ENV SOLR_UID 8983
 RUN groupadd -r -g $SOLR_UID $SOLR_USER && \
   useradd -r -u $SOLR_UID -g $SOLR_USER $SOLR_USER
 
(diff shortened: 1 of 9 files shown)


tianon:
Closing in favor of https://github.com/docker-solr/docker-solr/pull/114! :+1:

Issue #120, opened 2017-06-08
scripts not shellcheck-compliant


makuk66:
We should fix:

```
mak@cylon 2167 docker-solr [master] $ for f in scripts/*; do echo $f; shellcheck $f; done
scripts/docker-entrypoint.sh
scripts/init-solr-home

In scripts/init-solr-home line 42:
if [ $(find "$SOLR_HOME" -mindepth  1 | wc -l) != '0' ]; then
     ^-- SC2046: Quote this to prevent word splitting.

scripts/run-initdb

In scripts/run-initdb line 32:
done < <(ls /docker-entrypoint-initdb.d/ | sort -n)
         ^-- SC2012: Use find instead of ls to better handle non-alphanumeric filenames.

scripts/solr-create

In scripts/solr-create line 15:
echo "Executing $0 $@"
                   ^-- SC2145: Argument mixes string and array. Use * or separate argument.


In scripts/solr-create line 29:
    echo "Creating core with: ${@:1}"
                              ^-- SC2145: Argument mixes string and array. Use * or separate argument.


In scripts/solr-create line 39:
    echo "Created core with: ${@:1}"
                             ^-- SC2145: Argument mixes string and array. Use * or separate argument.

scripts/solr-demo
scripts/solr-foreground
scripts/solr-precreate

In scripts/solr-precreate line 14:
echo "Executing $0 $@"
                   ^-- SC2145: Argument mixes string and array. Use * or separate argument.


In scripts/solr-precreate line 32:
    cp -r $CONFIG_SOURCE/ $coredir
          ^-- SC2086: Double quote to prevent globbing and word splitting.
                          ^-- SC2086: Double quote to prevent globbing and word splitting.

scripts/start-local-solr
scripts/stop-local-solr
scripts/wait-for-solr.sh

In scripts/wait-for-solr.sh line 14:
  echo $@
       ^-- SC2068: Double quote array expansions, otherwise they're like $* and break on spaces.


In scripts/wait-for-solr.sh line 34:
  if (( $i == $max_try )); then
        ^-- SC2004: $/${} is unnecessary on arithmetic variables.
              ^-- SC2004: $/${} is unnecessary on arithmetic variables.
```

makuk66:
```
$ for f in scripts/*; do echo $f; shellcheck $f; done
scripts/docker-entrypoint.sh
scripts/init-solr-home
scripts/run-initdb
scripts/solr-create
scripts/solr-demo
scripts/solr-foreground
[...]

PR #121, opened 2017-06-08
fix issue 120: shellcheck compliance

makuk66:
Fix https://github.com/docker-solr/docker-solr/issues/120: make shell scripts shellcheck compliant


makuk66:
Done elsewhere

Issue #128, opened 2017-08-17
support arm64 architecture


longquan7:
I want to build this image on arm64 architecture,but I'm not sure that can be support arm64 architecture https://archive.apache.org/dist/lucene/solr/6.6.0/solr-6.6.0.tgz

makuk66:
Solr just requires Java. And docker-solr uses shell scripts, expecting bash and so on. So if you find yourself a debian+openjdk or apline+openjdk image for arm, modify the FROM line in the docker-solr Dockerfiles to that, and build, it might just work. I don't have an arm64 machine up at the moment, so I can't try it out.
Even if this works fine, it would be tricky to get to build on our Travis build infrastructure or the docker-library infrastructure, so it's probably not something we'd be able to incorporate into this repo.

longquan7:
@makuk66 OK,thanks

makuk66:
I created an arm64 node on [Scaleway](https://www.scaleway.com), based on their docker image.

Uname reports it as: 

```
Linux scw-c77bff 4.9.32 #1 SMP Wed Jun 14 18:37:07 UTC 2017 aarch64 aarch64 aarch64 GNU/Linux
```

I found an arm openjdk image for arm64 on [Docker hub](https://hub.docker.com/search/?isAutomated=0&isOfficial=0&page=1&pullCount=0&q=openjdk-arm&starCount=0).

```
$ docker pull arm64v8/openjdk

$ docker run arm64v8/openjdk java -version
openjdk version "1.8.0_141"
OpenJDK Runtime Environment (build 1.8.0_141-8u141-b15-1~deb9u1-b15)
OpenJDK 64-Bit Server VM (build 25.141-b15, mixed mode)

$ git clone https://github.com/docker-solr/docker-solr.git

$ cd docker-solr/6.6
$ sed -i 's,FROM .*,FROM arm64v8/openjdk,' Dockerfile
$ docker build -t solr-arm .
$ docker run -p 8983:8983 solr-arm solr-demo
```

That ran into some key fetching timeouts, but eventually built, and the demo runs fine. Neat.

Note: this exposes Solr on the public internet without protection; so firewall/proxy as appropriate.

longquan7:
@makuk66 That' great, So Solr whether or not support arm64 architecture, can we push it to the https://hub.docker.com/u/arm64v8/  ?

makuk66:
On a scaleway `ARM64-8GB`:

```
[...]

PR #131, opened 2017-08-21
Update generate-stackbrew-library.sh for multiple architecures

makuk66:
Update generate-stackbrew-library.sh to the latest from
https://github.com/docker-library/official-images/blob/master/library/elasticsearch
which uses RFC 2822 format.
Additional tweak to skip non-version directories.

Add other architectures supported by
https://github.com/docker-library/official-images/blob/master/library/openjdk

For https://github.com/docker-solr/docker-solr/issues/128

### DIFF
--- a/generate-stackbrew-library.sh
+++ b/generate-stackbrew-library.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Based on https://github.com/docker-library/elasticsearch/blob/master/generate-stackbrew-library.sh
+# Based on https://github.com/docker-library/httpd/blob/master/generate-stackbrew-library.sh
 set -eu
 
 declare -A aliases=(
@@ -10,6 +10,7 @@ declare -A aliases=(
 
 self="$(basename "$BASH_SOURCE")"
 cd "$(dirname "$(readlink -f "$BASH_SOURCE")")"
+
 extglob_old=$(shopt -p extglob||true)
 shopt -s extglob
 
@@ -43,6 +44,22 @@ dirCommit() {
 	)
 }
 
+getArches() {
+	local repo="$1"; shift
+	local officialImagesUrl='https://github.com/docker-library/official-images/raw/master/library/'
+
+	eval "declare -g -A parentRepoToArches=( $(
+		find -name 'Dockerfile' -exec awk '
+				toupper($1) == "FROM" && $2 !~ /^('"$repo"'|scratch|microsoft\/[^:]+)(:|$)/ {
+					print "'"$officialImagesUrl"'" $2
+				}
+			' '{}' + \
+			| sort -u \
+			| xargs bashbrew cat --format '[{{ .RepoName }}:{{ .TagName }}]="{{ join " " .TagEntry.Architectures }}"'
+	) )"
+}
+getArches 'solr'
+
 cat <<-EOH
 # this file is generated via https://github.com/docker-library/solr/blob/$(fileCommit "$self")/$self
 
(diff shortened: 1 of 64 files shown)


dsmiley:
Sorry I'm confused; what is this about?

makuk66:
This is support for architectures other than amd64. This was prompted by #128, where longquan7 was looking for an ARM version. Given that Solr is Java, we should be able to run on any architecture for which a suitable JDK image is available. I tested on arm64 and that seems to work; I don't have access to other architectures.

Docker hub has support for building for multiple architectures; see https://github.com/docker-library/official-images#multiple-architectures. To support that, we need to update the docker library manifest from the current:

```
6.4.2: git://github.com/docker-solr/docker-solr@e3a05300ab8e5f1dde548a9b3b3be79d23f2bb16 6.4
6.4: git://github.com/docker-solr/docker-solr@e3a05300ab8e5f1dde548a9b3b3be79d23f2bb16 6.4

6.4.2-alpine: git://github.com/docker-solr/docker-solr@e3a05300ab8e5f1dde548a9b3b3be79d23f2bb16 6.4/alpine
6.4-alpine: git://github.com/docker-solr/docker-solr@e3a05300ab8e5f1dde548a9b3b3be79d23f2bb16 6.4/alpine
```

to a slightly different-looking format that specifies supported architectures:

```
Tags: 6.6, 6, latest
Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, ppc64le, s390x
GitCommit: 600fd5c682b63cca400ef87fe7738d752f6ca0ed
Directory: 6.6

Tags: 6.6-alpine, 6-alpine, alpine
GitCommit: c61a0c9b012c7313c2b5d0d97ddc06693270b734
Directory: 6.6/alpine
```

That manifest is created by the `generate-stackbrew-library.sh` shell script, which 
@tianon has updated for e.g. https://github.com/docker-library/official-images/blob/master/library/elasticsearch; so we can use that.

AIUI we should end up with e.g. https://hub.docker.com/r/arm64v8/solr.

generate-stackbrew-library.sh
```diff
@@ -1,51 +1,101 @@
...
+	echo "${out#$sep}"
+}
 
 for version in "${versions[@]}"; do
-        if [ ! -f "$version/Dockerfile" ]; then
-          continue
-        fi
+	commit="$(dirCommit "$version")"
+
+	fullVersion="$(git show "$commit":"$version/Dockerfile" | awk '$1 == "ENV" && $2 == "ELASTICSEARCH_VERSION" { gsub(/~/, "-", $3); print $3; exit }')"
```
tianon: `ELASTICSEARCH_VERSION`? :smirk: (probably meant `SOLR_VERSION` here)

generate-stackbrew-library.sh
```diff
@@ -1,51 +1,101 @@
...
+		${aliases[$version]:-}
+	)
+
 	echo
-	for va in "${versionAliases[@]}"; do
-		echo "$va: ${url}@${commit} $version"
-	done
+	cat <<-EOE
+		Tags: $(join ', ' "${versionAliases[@]}")
+		Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, ppc64le, s390x
```
tianon: Just a note that you'll want to make sure to update this if the `Architectures` of the `openjdk:8-jre` image ever change, or if the `FROM` of `solr` ever changes. :+1:

(We usually get a little crazy and scrape that directly via the `bashbrew` tool as in https://github.com/docker-library/httpd/blob/d8a2ea0c867407d2ecd335ee7836b31019cd484d/generate-stackbrew-library.sh#L36-L50, but that's definitely not necessary.  Just making sure you're aware that you'll need to manually ensure this list is correct. :+1:)

makuk66: Oh that's neat. I'm very much not interested in manual verification and having to remember things. I'll update to that. That means the script has a bashbrew dependency now, but that's not a big deal.
Running the new script I see that in fact the list I had was incorrect; in particular the alpine images are only for amd64. But I see slim versions, so I guess I can do another variant.

makuk66:
With this, the manifest looks like https://gist.github.com/makuk66/c077364fb4fb3c1269db89aa1859631e

Issue #110, opened 2017-02-26
wait-for-solr.sh can not talk to remote servers


makuk66:
The `wait-for-solr.sh` script hardcodes the Solr URL to localhost, and thus cannot be used for waiting on other containers. We should make this configurable, and use named argument parsing while we're at it (but stay backwards compatible).

makuk66:
Done

PR #132, opened 2017-08-29
issue110: wait-for-solr.sh rewrite

makuk66:
Rewrite wait-for-solr.sh so that it can take options.
Add an option for the URL to check.

For https://github.com/docker-solr/docker-solr/issues/110


makuk66:
manual merge

PR #136, opened 2017-09-04
Improve --user invocation with INIT_SOLR_HOME

makuk66:
- change solr-create/solr-demo to not use sentinel files but look for the core directory
- change start-local-solr to write the PID to the logs directory if needed
- change start-local-solr to use SOLR_OPTS to limit the first JVM start to localhost,
  instead of writing a config file
- patch the bin/solr script to use `id` instead of `whoami` so it does not complain about missing /etc/password entries for the current user
- patch the bin/solr script to invoke lsof without complaining about missing /etc/password entries for the current user

  With these changes you can:

```
mkdir mysolrhome logs
docker run -it \
  --mount type=bind,source=$PWD/mysolrhome,destination=/mysolrhome \
  --mount type=bind,source=$PWD/logs,destination=/logs \
  -e VERBOSE=yes \
  -e SOLR_HOME=/mysolrhome \
  -e SOLR_LOGS_DIR=/logs \
  -e INIT_SOLR_HOME=yes \
  -u "$(id -u):$(id -g)" \
  solr solr-create -c books
```

### DIFF
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -46,6 +46,8 @@ RUN mkdir -p /opt/solr && \
   rm /opt/solr.tgz* && \
   rm -Rf /opt/solr/docs/ && \
   mkdir -p /opt/solr/server/solr/lib /opt/solr/server/solr/mycores && \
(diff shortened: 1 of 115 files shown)


makuk66:
Manual merge

PR #141, opened 2017-09-13
add support to create multiple cores using solr-precreate

hlavki:
Updated solr-precreate script to preserve original usage and add support to create multiple cores

### DIFF
--- a/6.6/scripts/solr-precreate
+++ b/6.6/scripts/solr-precreate
@@ -9,6 +9,8 @@
 # To create a core in a mounted directory:
 #      mkdir mycores; chown 8983:8983 mycores
 #      docker run -it --rm -P -v $PWD/mycores:/opt/solr/server/solr/mycores solr solr-precreate mycore
+# To create multiple cores:
+#      docker run -P -d solr solr-precreate mycore1 /path/to/core1 mycore2 /path/to/core2
 set -e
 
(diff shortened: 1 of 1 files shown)


makuk66:
Thanks for your PR! It helps seeing exactly what you're trying to achieve.

Couple of comments about it:

- I can't merge the PR as-is, because it only changes one of the scripts.
  Instead, make changes in ./scripts, then run tools/update.sh which will copy it to all other directories
- That -gettingstarted is not right
- The changes cause `shellcheck` errors; I want to stay compliant.
- There's something I don't like about the pairwise arguments. For one, if you want to specify multiple cores you're now forced to provide the core config dir. For another, if we want to introduce further arguments, it becomes a mess to keep track of.

How about this instead: we pull solr-precreate apart, so you could do:
```
docker run -it -p 8983:8983 docker-solr/docker-solr:6.6.1 \
  bash -c "precreate-core core1; \
  precreate-core core2 /opt/solr/server/solr/configsets/data_driven_schema_configs; \
  precreate-core core3; \
  solr-foreground"
```
That would give you what you need, right? If so I'll create an issue for that, and test out my implementation a bit more.

hlavki:
Thanks @makuk66, your suggestion sounds good. PR was only as inspiration so I can close it.

PR #153, opened 2017-10-18
Ignore lost+found when checking if SOLR_HOME is empty in init-solr-home

### LINKED ISSUE #152: Ignore lost+found when checking if SOLR_HOME is empty in init-solr-home
init-solr-home sensibly checks whether a directory is empty before attempting to copy /home/solr/server/solr into it.

This precaution causes trouble when the mounted volume is a Unix filesystem as they contain a lost+found directory used to put files that were misplaced by the filesystem.

morissm:
init-solr-home sensibly checks whether a directory is empty before attempting to copy /home/solr/server/solr into it.

This precaution causes trouble when the mounted volume is a Unix filesystem as they contain a lost+found directory used to put files that were misplaced by the filesystem.

Fixes #152

### DIFF
--- a/5.5/alpine/scripts/init-solr-home
+++ b/5.5/alpine/scripts/init-solr-home
@@ -39,7 +39,7 @@ if [ -f "$SOLR_HOME/solr.xml" ]; then
 fi
 
 # refuse to use non-empty directories, which are likely a misconfiguration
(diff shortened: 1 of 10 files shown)


makuk66:
Thanks!

Issue #6, opened 2016-01-04
Start problem in kubernetes


xrow:
Hi,

I have noticed that you can`t start the docker image in kubernetes, because kubernetes will set the following env vars:

```
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=solr-rrdd1
SOLR_PORT_8983_TCP_PROTO=tcp
SOLR_PORT_8983_TCP=tcp://10.254.219.195:8983
SOLR_SERVICE_HOST=10.254.219.195
SOLR_SERVICE_PORT_SOLR=8983
SOLR_PORT=tcp://10.254.219.195:8983
SOLR_PORT_8983_TCP_ADDR=10.254.219.195
SOLR_PORT_8983_TCP_PORT=8983
SOLR_SERVICE_PORT=8983
SOLR_USER=solr
SOLR_UID=8983
SOLR_KEY=25C13C58807E90004AB8597AE48025ED13E57FFC
SOLR_VERSION=5.4.0
SOLR_SHA256=84c0f04a23047946f54618a092d4510d88d7205a756b948208de9e5afb42f7cd
```

The /opt/solr/bin/solr bin will act on SOLR_PORT set by kubernetes and will crash since it will have the value tcp://10.254.219.195:8983. I sugest using the var SOLR_SERVICE_PORT_SOLR, if var is set.

Else you have to change the name of your server in kubernetes from solr to solr2 for example.

xrow:
Another workaround is to set a var SOLR_PORT and overwrite the value.
   
 - name: "SOLR_PORT"
   value: "8983"

makuk66:
From https://github.com/kubernetes/kubernetes/blob/master/docs/user-guide/services.md#environment-variables it sounds like that is expected behaviour on the kubernetes side.

Your workaround sounds straightforward, and it seems that if you want it you can constitute the original value from SOLR_PORT_8983_TCP_ADDR and SOLR_PORT_8983_TCP_PORT if you were to need it.

It'd be nice if we could avoid this pitfall. I can think of a few approaches to avoid this:
- document your overwrite workaround in the https://github.com/docker-solr/docker-solr/blob/master/Docker-FAQ.md as part of a "how to run docker-solr in kubernetes" entry
- use a wrapper to call the solr start script, and unset SOLR_PORT explicitly. But so far we've avoided the need for a wrapper
[...]

PR #159, opened 2018-01-05
Make internal solr port configurable

tboeghk:
This PR makes the internal `SOLR_PORT` configurable using a Docker environment variable. 

I noticed that we do not need to explicitly configure the `SOLR_PORT` in the `solr.in.sh` during Docker build if we supply the `SOLR_PORT` as default Docker environment variable. This furthermore makes the internal `SOLR_PORT` configurable, which is a useful feature.

It is useful in 3rd party applications when running local integration tests with randomized ports in Solr cloud mode. We do this a lot using a local docker-compose network. In Docker Compose networks, Solr connects to the Zookeeper using it’s internal node name and internal port number. 

When connecting to Zookeeper using a Java client (SolrJ), the live nodes are read from Zookeeper. If you run Solr with randomized external Docker ports, you won't be able to connecto to them as Solr registered itself using it's internal port number. You can fix this with this PR and configuring both external and internal Docker port to the same number:

```
- ports:
    "12345:12345"
```
We'd love to see this PR accepted as this causes a lot of pain during our integration tests. This PR is building at Travis CI: https://www.travis-ci.org/shopping24/docker-solr


makuk66:
The approach here looks sensible and convenient. You missed the EXPOSEs in the templates (and we might as well stick with the {}-less syntax for consistency) and didn't run the `tools/update.sh`. You modified `.gitignore` which is not needed, but is harmless. You added `joe` and `less` to one `Dockerfile`, which are not required. It would be nice to have a test-case. So apply https://gist.github.com/makuk66/96702ff43fbeb9dfe3f74ae84eaf7dde to your branch and give that a go.

I can't off-hand see a backwards compatibility issue here, given that any `solr.in.sh` `SOLR_PORT` changes with initdb or mounting would fail (in the wait scripts) today.

tboeghk:
Uhpsi, `joe` and `less` should not have been there, my bad. I'll apply your patch asap. Thanks man!

tboeghk:
Unfortunately your diff was corrupt (line 130) but I did not find the problem. I rebuilt the changes myself, I hope I did not miss anything

makuk66:
Just `git update-index --chmod=+x tests/precreate_core_solr_port/test.sh`

tboeghk:
Hmmm, dunno why Travis did not pick up the commit above, but it's building on the shopping24 branch: https://www.travis-ci.org/shopping24/docker-solr/builds/325444471

makuk66:
Hmm, I had a feeling that SOLR_PORT sed was there for a reason: https://github.com/docker-solr/docker-solr/issues/6. We'll have to ponder that a bit more.

dsmiley:
Just checking but was the underlying problem fixed here the hard-coding of 8983 in some of the scripts?  I can see that's a problem.  Just reading the PR description I was a bit confused as to the problem since I can already set whatever env var I want when I invoke docker (so long as it's actually used by Solr and Solr doesn't override it).

makuk66:
Currently you cannot override SOLR_PORT by passing it in at docker run time, because we at docker build time patch `solr.in.sh` to set SOLR_PORT=8983 explicitly (this was a fix for https://github.com/docker-solr/docker-solr/issues/6 but perhaps not a very good one), and because even if you could set SOLR_PORT, the docker-solr wait scripts didn't pay attention.

The proposed change in this PR is three-fold:

- 1: set the SOLR_PORT env var to a default 8983 in the Dockerfiles. You can still override at runtime
- 2: remove the explicit adding of SOLR_PORT=8983 to the `solr.in.sh` at "docker build" time 
- 3: update the scripts to use the SOLR_PORT env var.

This works if you just run with defaults, and works if you explicitly pass SOLR_PORT as an env variable. Question is, what happens with kubernetes if you do not explicitly pass SOLR_PORT: does it override the Dockerfile one with its own invalid one? I can't test that easily at the moment but I'd like to. If the answer is yes, then it'd be nice to deal with that. Perhaps:

- do nothing, and let Solr fail, which seems unfriendly, even if we provide docs that point this out
- at docker build time add code to `solr.in.sh` to check for the invalid SOLR_PORT value and unset it. Ie do what we do now but be smarter about it and not trample of user-provided environment. Then we'd have to call that from the scripts too. Stil don't love it, because users may be mounting this.
- do this in the initdb instead. If that works for everything, that might be nice. Would need to try

But if the kubernetes case is not actually a problem after this patch, then we don't need to do anything which would be best. So I'd like that tried first.

tboeghk:
I can check on the Kubernetes case on monday.

tboeghk:
Sped up time a little and fired up a local `minikube` and deployed Solr/Zookeeper ...

I can confirm, that this PR still works when running Solr in Kubernetes. The internal `SOLR_PORT` can be changed and it does not get overwritten by Kubernetes. Kubernetes on the other hand does not interfere with the `SOLR_PORT` environment variable.

### Kubernetes setup

This is the very basic deployment I used. It deploys a Solr and a Zookeeper instance as a Kubernetes pod. It pulls a Solr Docker image built from this PR from the @shopping24 internal Docker hub and deploys it into k8s.

```yaml
apiVersion: apps/v1beta1
kind: Deployment
metadata:
  name: solr-deployment
spec:
  replicas: 1
  template:
    metadata:
      labels:
        app: solr
    spec:
      containers:
      - name: solr
        image: our-internal-docker-hub.s24.com/s24/solr:7.2.0-pr159
        env:
          - name: ZK_HOST
            value: localhost
          - name: SOLR_PORT
            value: "8123"
        ports:
        - containerPort: 8123
      - name: zookeeper
        image: zookeeper:3.4.11
        ports:
        - containerPort: 2181
      imagePullSecrets:
      - name: our-internal-pull-secret
```

I deployed this twice, first without the `SOLR_PORT` being configured, then as shown above. I exposed the deployment (the rough and simple way) using `kubectl expose deployment solr-deployment --type=NodePort` and checked the Solr web interface using `minikube service solr-deployment`.

### First run, no `SOLR_PORT` configured

This is the Solr environment before it had been exposed as a service. There are no environment variables set by Kubernetes:

01_solr_k8s_environment_unexposed

After being exposed as a service, this environment is set in the Solr container. The default `SOLR_PORT` variable is not overwritten.

04_solr_k8s_environment_exposed_default_solr_port

The default port is used when registering the node in Zookeeper (as expected):

03_solr_k8s_default_port_in_zookeeper

### Second run, `SOLR_PORT` set to `8123`

After being exposed as a Kubernetes service, this environment is set to the Solr container. The explicitly set `SOLR_PORT` variable is not overwritten:

05_solr_k8s_environment_exposed_explicit_solr_port

The configured port `8123` is used when registering the node in Zookeeper (as expected):

06_solr_k8s_explicit_port_in_zookeeper

### Conclusion

To me it seems that `Dockerfile` environment variables and Kubernetes variables take precedence over variables set by Kubernetes service exposure. -double checks appreciated. This PR removes the hard codes `SOLR_PORT` from scripts and makes it more configurable via environment. This has the potential to fix Docker and Docker Compose setups using randomized ports (integration tests) and leaves Kubernetes deployments untouched.

makuk66:
Thanks for that write-up, that helped me get started. But, there is a problem. You used `solr-deployment` as the app name, so the variables ended up being `SOLR_DEPLOYMENT_PORT` which don't conflict, so this is not an accurate repro of #6. Thus the conflict not occurring on the branch doesn't tell us anything.

Here's what I did: checkout master, remove the SOLR_PORT sed from the Dockerfile templates, rebuild, push as `makuk66/solrport:broken`. Then use (note name `solr`):

`solr-deployment-broken.yaml`:
```
apiVersion: apps/v1beta1
kind: Deployment
metadata:
  name: solr
spec:
  replicas: 1
  template:
    metadata:
      labels:
        app: solr
    spec:
      containers:
      - name: solr
        image: index.docker.io/makuk66/solrport:broken
        env:
          - name: ZK_HOST
            value: localhost
        ports:
        - containerPort: 8983
      - name: zookeeper
        image: zookeeper:3.4.11
        ports:
        - containerPort: 2181
```

`show-env.yaml`:
```
apiVersion: v1
kind: Pod
metadata:
  name: show-env
spec:
  containers:
  - name: show-env
    image: nginx
    command: [ "env" ]
  restartPolicy: Never
```
Then run:
```
kubectl create -f solr-deployment-broken.yaml
kubectl expose deployment solr --type=NodePort
kubectl create -f show-env.yaml
sleep 2
kubectl logs show-env|sort|grep SOLR
```
that shows `SOLR_PORT=tcp://10.97.34.254:8983`

So now:
```
kubectl delete pod show-env
kubectl delete service solr
kubectl delete -f solr-deployment-broken.yaml
```
modify the image to `index.docker.io/makuk66/solrport:7.2.0-pr159` (your branch).

repeat the steps, and I still see `SOLR_PORT=tcp://10.105.209.125:8983`.

Add the SOLR_PORT env:

```
          - name: ZK_HOST
            value: localhost
          - name: SOLR_PORT
            value: "7777"
```
repeat the steps, and I still see `SOLR_PORT=tcp://10.99.2.245:8983`.

So, I think this branch would break #6 again, and we'll have to look at some more logic in our scripts.

makuk66:
On reflection, my previous doesn't entirely make sense either (need to actually look at the solr container too). I'm gonna have to get back to this.

makuk66:
[BTW, next time please create an issue before creating a PR; that makes it easier to compare different PRs as potential fixes for the issue.]

As I can't write to your branch, I created a new branch and PR: https://github.com/docker-solr/docker-solr/pull/163
There I added logic to the scripts to fail with an appropriate error, and added logic to the docker entrypoint to fix up the `SOLR_PORT` in the kubernetes scenario.

To test, I did:

1. Build the branch:

```
bash tools/build_test_all.sh
```

2. Tag and push to a test repo, from where kubernetes can pull it:

```
docker tag docker-solr/docker-solr:latest makuk66/solrport:take3
docker push makuk66/solrport:take3
```

3. Create a solr deployment named "solr" and expose as a service

```
cat > solr-deployment-solr-take3-1.yaml <<EOM
apiVersion: apps/v1beta1
kind: Deployment
metadata:
  name: solr
spec:
  replicas: 1
  template:
    metadata:
      labels:
        app: solr
    spec:
      containers:
      - name: solr
        image: "makuk66/solrport:take3"
        env:
          - name: ZK_HOST
            value: localhost
          - name: VERBOSE
            value: "yes"
        ports:
        - containerPort: 8983
      - name: zookeeper
        image: zookeeper:3.4.11
        ports:
        - containerPort: 2181
EOM
kubectl create -f solr-deployment-solr-take3-1.yaml
kubectl rollout status deployment solr
kubectl expose deployment solr --type=NodePort
```

Kubernetes will now set `SOLR_PORT=tcp://something:8983` in future containers.

4. Create a second solr deployment:

```
cat > solr-deployment-take3-noport.yaml <<EOM
apiVersion: apps/v1beta1
kind: Deployment
metadata:
  name: solr-take3-noport
spec:
  replicas: 1
  template:
    metadata:
      labels:
        app: solr
    spec:
      containers:
      - name: solr
        image: "makuk66/solrport:take3"
        env:
          - name: ZK_HOST
            value: localhost
          - name: VERBOSE
            value: "yes"
        ports:
        - containerPort: 8983
      - name: zookeeper
        image: zookeeper:3.4.11
        ports:
        - containerPort: 2181
EOM
kubectl create -f solr-deployment-take3-noport.yaml
kubectl rollout status deployment solr-take3-noport
kubectl get pods|grep solr-take3-noport
kubectl logs logs solr-take3-noport-5b66d947-wr9sm solr
```

Kubernetes set `SOLR_PORT` to its "bad" value, but `docker-entrypoint.sh` fixed it up:

```
SOLR_PORT=tcp://10.101.164.27:8983
...
Invalid SOLR_PORT=tcp://10.101.164.27:8983 environment variable specified; resetting to 8983
+ echo 'Invalid SOLR_PORT=tcp://10.101.164.27:8983 environment variable specified; resetting to 8983'
+ SOLR_PORT=8983
+ export SOLR_PORT
...
2018-01-16 13:06:21.354 INFO  (main) [   ] o.a.s.s.SolrDispatchFilter / __| ___| |_ _   Starting in cloud mode on port 8983
```
5. create a solr deployment named "solr-take3-port7777", with `SOLR_PORT=7777` configured:

```
cat > solr-deployment-take3-port7777.yaml <<EOM
apiVersion: apps/v1beta1
kind: Deployment
metadata:
  name: solr-take3-port7777
spec:
  replicas: 1
  template:
    metadata:
      labels:
        app: solr
    spec:
      containers:
      - name: solr
        image: "makuk66/solrport:take3"
        env:
          - name: ZK_HOST
            value: localhost
          - name: VERBOSE
            value: "yes"
          - name: SOLR_PORT
            value: "7777"
        ports:
        - containerPort: 7777
      - name: zookeeper
        image: zookeeper:3.4.11
        ports:
        - containerPort: 2181
EOM
kubectl create -f solr-deployment-take3-port7777.yaml
kubectl get pods|grep solr-take3-port7777
```

Kubernetes set `SOLR_PORT=7777`, and Solr started there:

```
2018-01-16 13:09:46.678 INFO  (main) [   ] o.a.s.s.SolrDispatchFilter / __| ___| |_ _   Starting in cloud mode on port 7777
```

So, that all looks good. It's conceivable I've missed something though; comments appreciated.
One question that is debatable: should we printing out that `Invalid SOLR_PORT` line? It might be of interest for normal logs, but it might get in the way of other situations where the users uses the image for something other than running solr. And if you want to know if it's happening you can run with `VERBOSE=yes`. So I think I'll remove it.

makuk66:
Of course, if we go this way, we don't actually need to set SOLR_PORT in the Dockerfile at all; we can leave it unset unless it is specifically required.
@tboeghk can you test if https://github.com/docker-solr/docker-solr/pull/163 works for you?

makuk66:
Merged https://github.com/docker-solr/docker-solr/pull/163 in favour of this one

PR #160, opened 2018-01-05
Add centos based Solr images

tboeghk:
In addition to `alpine`, `slim` and `debian` based distros this adds a `centos` based Docker-Solr image. This adds a new template and adjusts tests to handle centos distros as well.

As there is no ready made openjdk-centos image, a recent jdk is installed in this Dockerfile. Moreover the Java version can be specified via environment variable. Also the centos base image can be changed as Docker build argument. This can be handy, if you need a specific centos based image that uses the `yum` package manager.

I'm not sure if I missed a spot to adjust, but [Travis is ok with it](https://www.travis-ci.org/shopping24/docker-solr/builds/325456672).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ upstream_versions
 tests/tests_to_run
 mycores
 scripts.old
+/.project
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,10 @@ jobs:
       script: tools/build_test_push.sh 7.2/alpine
     - stage: build, test, deploy
       env:
+      - PROCESS=7.2/centos
+      script: tools/build_test_push.sh 7.2/centos
+    - stage: build, test, deploy
+      env:
       - PROCESS=7.2
       script: tools/build_test_push.sh 7.2
     - stage: build, test, deploy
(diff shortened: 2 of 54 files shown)


dsmiley:
I have to wonder if there is a failure of Docker itself if every project feels the need to have a matrix of images for every OS + Java variant (or other language runtime e.g. ruby, python).  It's extra work to maintain these combinations, and produces lots of extra images (waste).  This isn't a direct vote against for this PR... I'm more or less me ruminating on the container world right now.  It's sad to see this.

tboeghk:
@dsmiley I see your point, but let me explain. 

I maintain a bunch of `docker-solr` installations for different companies. While some companies don't care about the distro used, most demand a RedHat/CentOS based os in their Docker images (at least in my filter bubble).

And this happens for a reason. All companies extend the `docker-solr` image and add custom os packages regarding Java inspection, logging, profiler and some Java custom code, Solr plugins etc. For those os packages most  companies have RPMs ready and they are built to run on `yum` based systems. 

Until now for each installation I forked this repo into a private one, changed the build system from `apt-get` to `yum` and published the Docker images into a companies private Docker hub. So before doing this the nth time over I thought it might be a good idea to open source this. The main idea is to have a `docker-solr` image using `yum` as a package manager.

makuk66:
I agree with @dsmiley here. We're based off official Docker library `openjdk` images, and the only reason we have alpine is for size, and now slim has taken over that role. I do not want to go beyond that unless upstream do.

tboeghk:
I'm closing this PR then. To not lose the time and effort invested, I pushed the images to Docker Hub and will be maintaining them myself: https://hub.docker.com/r/shopping24/docker-solr-centos/

PR #161, opened 2018-01-08
Prepare docker-solr-centos fork

tboeghk:


tboeghk:
Uhpsi, wrong master repo

PR #176, opened 2018-05-31
Allow persistence storage in orchestrators

### LINKED ISSUE #175: Mounting Volumes over Solr Owned Directories Causes Permission Errors
If I would like to have persistence storage in an orchestrator such as DC/OS, I need to ensure that the SOLR_HOME directory can be written to by the Solr user and this is currently not possible. When mounting a volume over a Solr owned directory, the directory becomes owned by root. 

In order to resolve this, I believe that the SOLR_HOME directory should have it's ownership changed during the docker-entrypoint.db as this is the earliest point in which we can ensure that when Solr writes to this directory, it is able to. 

I believe there needs to be a few changes in the `Dockerfile` and in the `docker-entrypoint.sh` in order to allow this:
1. Ensure that the `Dockerfile` doesn't set the user to `solr`
1. Ensure that we have sudo available within the Solr image
1. Create the `$SOLR_HOME` directory and change ownership to solr:solr in the `docker-entrypoint.sh`
1. Change the exec command in `docker-entrypoint.sh` to run the command as the solr user, whilst maintaining environment variables and PATH

corbs9:
- Ensured that the SOLR_HOME directory is writable when it is mounted as a volume
Closes #175

### DIFF
--- a/7.3/Dockerfile
+++ b/7.3/Dockerfile
@@ -1,4 +1,3 @@
-
 FROM openjdk:9-jre
 MAINTAINER  Martijn Koster "mak-docker@greenhills.co.uk"
 
@@ -7,7 +6,11 @@ MAINTAINER  Martijn Koster "mak-docker@greenhills.co.uk"
 ARG SOLR_DOWNLOAD_SERVER
 
 RUN apt-get update && \
-  apt-get -y install lsof procps wget gpg && \
+  apt-get -y install lsof \
+  procps \
+  wget \
+  sudo \
+  gpg && \
   rm -rf /var/lib/apt/lists/*
 
 ENV SOLR_USER="solr" \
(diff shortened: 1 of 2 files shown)


corbs9:
At present, this is only done for the 7.3 image, but I can work on the other versions if you think this is good enough.

makuk66:
See my response in #175; let's discuss there. I believe you may be able to achieve your aims without requiring any changes to the docker-solr image.

Comments on this PR:
- The addition of sudo is not needed because we have [gosu](https://github.com/tianon/gosu).
- Whether the "run as root then chown everything" is a good idea is less clear at the moment, but is being thought about

> At present, this is only done for the 7.3 image, but I can work on the other versions

The way to make changes is to make then in the `./scripts` directory and in the `*template` files, then run `bash tools/update.sh` to apply them to all the versions. See https://github.com/docker-solr/docker-solr/blob/master/update.md

makuk66:
BTW, you also might find https://github.com/docker-solr/docker-solr/compare/chown-and-run of interest, in particular https://github.com/docker-solr/docker-solr/compare/chown-and-run#diff-4e421a15cb9cc2176b9e86f4d2ae01d8R16 . This was an experimental branch, which may be incomplete, and I'm not sure is necessarily a good idea or worth it.

PR #187, opened 2018-10-03
Add optional build arg ADD_JDBC_PSQL_DRIVER to auto download postgreSQL Driver

gadelkareem:
`docker build -t my_solr . --build-arg ADD_JDBC_PSQL_DRIVER=1` builds solr with jdbc.postgresql.org driver

### DIFF
--- a/7.5/Dockerfile
+++ b/7.5/Dockerfile
@@ -6,6 +6,10 @@ MAINTAINER  Martijn Koster "mak-docker@greenhills.co.uk"
 #   docker build -t mine --build-arg SOLR_DOWNLOAD_SERVER=http://www-eu.apache.org/dist/lucene/solr .
 ARG SOLR_DOWNLOAD_SERVER
 
+ARG ADD_JDBC_PSQL_DRIVER
+
+ENV JDBC_PSQL_VERSION 42.2.5
+
 RUN apt-get update && \
   apt-get -y install lsof procps wget gpg && \
   rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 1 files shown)


makuk66:
Hi Gadel,

This PR failed to build on CI. I can't push to your branch, but this does seem to work locally:

```diff --git a/7.5/Dockerfile b/7.5/Dockerfile
index 88896ed..dcdca56 100644
--- a/7.5/Dockerfile
+++ b/7.5/Dockerfile
@@ -78,9 +78,10 @@ RUN set -e; \
 COPY scripts /opt/docker-solr/scripts
 RUN chown -R $SOLR_USER:$SOLR_GROUP /opt/docker-solr
 
-RUN [ ! -z "$ADD_JDBC_PSQL_DRIVER" ] \
-    && mkdir /opt/solr/contrib/dataimporthandler/lib \
-    && wget -O /opt/solr/contrib/dataimporthandler/lib/postgresql-$JDBC_PSQL_VERSION.jar "http://jdbc.postgresql.org/download/postgresql-$JDBC_PSQL_VERSION.jar"
+RUN set -e; if [ ! -z "$ADD_JDBC_PSQL_DRIVER" ]; then \
+      mkdir -p /opt/solr/contrib/dataimporthandler/lib; \
+      wget -O "/opt/solr/contrib/dataimporthandler/lib/postgresql-$JDBC_PSQL_VERSION.jar" "http://jdbc.postgresql.org/download/postgresql-$JDBC_PSQL_VERSION.jar"; \
+    fi
 
 EXPOSE 8983
 WORKDIR /opt/solr
```
But that was only for the 7.5 Dockerfile; you should add it to the `Dockerfile.*template` files, then run `tools/update.sh` to copy it everywhere.

However, I have a more general objection to this approach of adding plugins to the base Solr image:
- it's a slippery slope (why not a different JDBC version, or the mysql jdbc connector, or any other plugin).
- it's tricky to track versions of such components (when do we check for updates, how do we verify that the new version works)
- you're adding something that can only be used if you provide a special build argument, so that's not tested by our build infrastructure (and if we were to add it by default, we'd have to do a lot more testing).

Alternative approaches are:
- use your own Dockerfiles that inherit Solr, to build an image that has exactly what you want
- use a initdb script that downloads this requirement at docker `run` time
- use docker compose and mounted named volumes to put it in place from one container you maintain

gadelkareem:
I agree, closing this...

Issue #173, opened 2018-05-16
Leverage the service installer script included with Solr


elyograg:
For quite some time now, Solr has included a service installer script.

Looking into the filesystem layout of users who ask for help on the solr-user mailing list with docker images, I have concluded that the layouts, configurations, and scripts were not created by that installer script, but were custom-written.

It would be a lot easier to support users on the solr-user mailing list when they're running docker images if the system layout were more in line with what we expect from a service installation.

Info that may already be known:

The service installer script creates a user (if it doesn't already exist), an init script, a configuration script in /etc/default, and sets the solr home to a location outside of the install directory.

https://lucene.apache.org/solr/guide/7_3/taking-solr-to-production.html#run-the-solr-installation-script

If the service name is set to something custom, such as solr7, then that affects the files and directories that are created.  The service name defaults to solr.  For a service name of solr7, here's what will exist after the service is started:

/opt/solr-X.X.X
/opt/solr7 (a symlink to the directory above)
/var/solr7
/var/solr7/data (this is the solr home)
/var/solr7/logs
/var/solr7/log4j.properties
/var/solr7/solr-8983.pid
/etc/default/solr7.in.sh
/etc/init.d/solr7

Everything in /opt/solr-X.X.X is owned by root.  The regular user (default solr) has no need to write to it.

Many settings can be configured by editing the include script in /etc/default.

A service installed this way can be upgraded by running the new installer with the -f option and all the options used before.  When that is done, the symlink is changed to point to the new version.

Things can get tricky when using the bin/solr script if the service name is not "solr", unless there's an environment variable pointing at the include script.  I'm hoping to improve the scripts so that installed services are found and there's a way to choose one if there are multiple.

https://issues.apache.org/jira/browse/SOLR-9590

Disclaimer: I'm not at all familiar with docker itself, or how you might need to customize things beyond what the installer does in order to accommodate usage in docker.  If you decide to use what we've provided and find problems in the scripts that cause issues for your project, we welcome bug reports, with or without patches.

makuk66:
> For quite some time now, Solr has included a service installer script.
> Looking into the filesystem layout of users who ask for help on the solr-user mailing list with docker images, I have concluded that the layouts, configurations, and scripts were not created by that installer script, but were custom-written.

[...]

PR #207, opened 2019-03-18
Solr8 installer

makuk66:
Add Solr 8.
Changes: http://lucene.apache.org/solr/8_0_0/changes/Changes.html
Being a major release, Solr 8 removes many deprecated APIs, changes various parameter defaults and behavior. Some changes may require a re-index of your content. You are thus encouraged to thoroughly read the "Upgrade Notes". Solr 8.0 also includes many other new features as well as numerous optimizations and bugfixes of the corresponding Apache Lucene release.

On the docker-solr side, switch the on-disk layout for 8 and beyond: instead of the expanded Solr tar, use the `install_solr_service` script. Fixes https://github.com/docker-solr/docker-solr/issues/173

temporary test images:
```
dockersolr/docker-solr:8.0.0-82a2f85
dockersolr/docker-solr:8.0.0-alpine-82a2f85
dockersolr/docker-solr:8.0.0-slim-82a2f85
```


dsmiley:
I'm all for using the install_solr_service to thus be more standard with upstream.

Why did the changes here touch previous releases?  Wouldn't we want them to stay frozen as they were?  I know with most new releases (all?) you touch old ones too but I don't understand why.

risdenk:
FWIW did some quick Solr S3 testing here based on this image:
https://github.com/risdenk/solr-s3a-testing

makuk66:
> Why did the changes here touch previous releases? Wouldn't we want them to stay frozen as they were? I know with most new releases (all?) you touch old ones too but I don't understand why.

For practicality really. When I update things in Dockerfiles or scripts, they are typically fixes or improvements, and I'd like those to be available to all 16 versions that the Docker Library team is building. To ensure that happens consistently, I use the `tools/update.sh` script which renders the Dockerfile templates and copies the scripts to all version and variant directories. If I didn't have that, I'd have to manually make those changes, and would be more likely to make mistakes. This has been working really well, when we've had one consistent way in which the scripts work. Now that there is a different way for 8, versus prior versions, it's a little more complicated, but by expressing the differences in a checked-in script, it is easier to understand that they are intentional, and they are easier to manage, than if I had to remember manual changes. We could implement that differently, say in this case have separate `scripts/old-style` and `scripts/service-installer` directories, and the same for the test directories, but then you end up with duplicated files, which risk becoming out of sync or out of date, and that becomes unwieldy.

We could alternatively throw out all older versions, so they do not get rebuilt (but remain on Docker hub), and we don't have to worry about maintaining them; but that would mean users of those versions would not receive security fixes and other improvements in the base images.

Basically, if you want to absolutely stick to un-changing bits-in-the-image for a specific installation, for the ultimate in repeatability, you should tag a specific image, and then host that yourself in a private registry. Then you can update/test/deploy that explicitly, at a schedule that suits you.

But yeah, it always looks scary when there are 162 files changed. When reviewing, you're best off to look changes in the templates, the `scripts/` and the `tools/`, and then just scan over the copies in the versions/variants.

dsmiley:
The main risk I see is testing to ensure we don't introduce a regression.  How confident are we that our generic scripts don't break some peculiarity of 5.x?  Especially with minutia like the ability to use an environment variable to override a default value.  I don't know.  Have we screwed up before?

makuk66:
> The main risk I see is testing to ensure we don't introduce a regression. How confident are we that our generic scripts don't break some peculiarity of 5.x? Especially with minutia like the ability to use an environment variable to override a default value. I don't know. Have we screwed up before?

When change happens you can't rule out regressions (and Solr has its own too); which is why I say that if zero regression is crucial to you, you should tag and host images that you have vetted, and that applies to all docker images. Like any software project, if you do make changes, a regression can happen, and then you have to deal with it, either by changing the software, or how it is used. We have some minimal testing to spot major issues, which should help some.

The question is, what do you do. I have explored the alternative approach I mentioned, see https://github.com/docker-solr/docker-solr/compare/solr-8-installer-copied2?expand=1 . There we have no changes to the pre-8 files, at the cost of duplication. I think that's fine too.

dsmiley:
Personally I prefer #209 since there is a notable change in this major release in where we put the files, and I don't think we should disturb the previous releases.  That said, I sympathize with your arguments and won't argue it any further.  I wonder what the "normal" / typical approach to this is on other projects?  I dunno; just thinking aloud.

Any way, thanks for doing all this work!

risdenk:
@makuk66 curious what the status is here? Was surprised to see that 8.0.0 wasn't released yet. Looks like might have been held up by 6.6.6 release?

Let me know if there is anything I can help with.

makuk66:
During further testing I discovered that Solr's `install_solr_service.sh` causes us problems:

- it does not support Alpine
- it does init stuff that is not relevant to containers
- it ignores errors

The [OS detection](https://github.com/apache/lucene-solr/blob/master/solr/bin/install_solr_service.sh#L62) tries to determine the OS in various ways until it sees something it recognises. When running on an unknown distro under Docker, it ends up looking at the running kernel (which is the host's), and that may confuse it. For example, when running an Alpine container on an Ubuntu host, it will spot the "Ubuntu" compiler in the kernel version line, and conclude wrongly it is running on Ubuntu. Whereas when running on Docker on Mac, an Alpine container runs on a Alpine VM, the kernel line only contains Alpine, and you get an "Unable to auto-detect" error.

The [auto-start service](https://github.com/apache/lucene-solr/blob/master/solr/bin/install_solr_service.sh#L354) is not something we want in a container. The command fails, because the container do not have that command. But the script ignores errors, and just carries on, so it kinda works, but it's a bit of a fudge.

The script also allows for various overrides, and different dist files and so on, which are not needed for our use-case.

So I am now not all that keen on using it. I'm bummed about that, because I figured this was done, and it has cost me enough time already. The alternative is to doing an install the same way but without the actual script, and then we can simply put it in the Dockerfile for clarity and see what is going on. Working through that took a while, but seems to be in a reasonable place now, see https://github.com/docker-solr/docker-solr/pull/210

risdenk:
@makuk66 ah that is frustrating. I didn't test the Alpine integration. Thanks for the details.

PR #210, opened 2019-04-10
Add Solr 8, using /var/solr

makuk66:
Add Solr 8.0
=========
This version of docker-solr has a changed on-disk layout, to bring it more in-line with the [Taking Solr to Production](https://lucene.apache.org/solr/guide/7_7/taking-solr-to-production.html) and the default layout produced by `install_solr_service.sh`. This makes it easier to use volumes to store your data.

Upgrade notes
---------------
- If you do any mounting of files or directories in your deployment, or run scripts that make modifications at runtime, you will need to review the changes documented below, and make appropriate modifications.
- If you are not ready to move to this new layout, stick to `solr:7`/`solr:7-slim`/`solr:7-alpine` rather than `solr`/`solr:latest`/`solr:slim`/`solr:alpine`. As always, if you want to guarantee absolutely no changes, keep a tagged image in your private repository, or build your own from source.
- This is the first release with this layout. It is conceivable further tweaks will be made to fix issues arising, so use `solr:8.0.0`/`solr:8.0.0-slim`/`solr:8.0.0-alpine` and explicitly check for changes before updating to later versions. Any such changes will be documented in the [solr PRs for the official images library](https://github.com/docker-library/official-images/pulls?utf8=✓&q=is%3Apr+solr) and in commit messages in [this repo](https://github.com/docker-solr/docker-solr/commits/master).

Layout Changes
----------------
- The Solr distribution is stored in `/opt/solr-$VERSION`, with `/opt/solr` symlinked to it. These files are root-owned, and can't be written to by Solr, which is desirable for security.
- The data is stored in `/var/solr/data`.
- The logs are in `/var/solr/logs`.
- The configuration is in `/etc/default/solr.in.sh`. I don't think having this file in `/etc/default` is all that useful over just leaving it where it is, as in practice you could mount over it in either place, but we'll follow `install_solr_service.sh` here. The `SOLR_INCLUDE`/`SOLR_HOME`/`SOLR_PID_DIR`/`SOLR_LOGS_DIR`/`LOG4J_PROPS` environment variables are declared in the `Dockerfile`s rather than in written `/etc/default/solr.in.sh`, for clarity and to allow other scripts to use those values, and so that you can override the values at run-time should you need to.
- The log configuration is in `/var/solr/log4j2.xml`. I'm not wild about a fixed configuration file living in `/var`, but that's where `install_solr_service.sh` puts it. You can override `LOG4J_PROPS` at runtime if want to, and point it at `/opt/solr/server/resources/log4j2.xml` or elsewhere.
- I made `/opt/docker-solr` root-owned too. I had to modify one of the tests accordingly.
- If you mount an empty volume onto `/var/solr`, the image will populate it with the `data` and `logs` directory, and the `solr.xml`, `zoo.cfg` and `log4j2.xml` files so that Solr will start.
- The `/var/solr` directory is marked as a VOLUME in the Dockerfile, so that even if you do not mount a volume there, Docker will create an anonymous volume there, which improves performance.
- The `/var/solr` directory has root group ownership, so the image works out of the box on OpenShift

Feature Removals
-----------------
- there is no longer a `mycores` directory. Mount volumes onto `/var/solr` instead.
- The `INIT_SOLR_HOME` mechanism has been removed. With the new separation of code and data I believe it should no longer be necessary; and the runtime preparation of the required directories and files in an empty `/var/solr` is now the default. 

Notes
------
- Why not just run `install_solr_service.sh` itself? Because it does not support Apline, installs init scripts we do not want, ignores failures, hardcodes configuration in `solr.in.sh`, ignores errors, and makes it hard to see what it's actually doing.


makuk66:
Test images:
- `dockersolr/docker-solr:8.0.0-876ec3dad0bf4151781e97bde54fd3468ac75240`
- `dockersolr/docker-solr:8.0.0-slim-876ec3dad0bf4151781e97bde54fd3468ac75240`
- `dockersolr/docker-solr:8.0.0-alpine-876ec3dad0bf4151781e97bde54fd3468ac75240`

dsmiley:
Overall looks great!  Thanks for your tireless work!

> there is no longer a mycores directory

That in particular makes me very happy :-). It's presence made the Docker-ization unclean in my eyes; I always wanted it to be as close as possible to upstream.  And in multiple ways, we're closer now than before; perhaps perfection.

p.s. I'm on vacation and as-such not very responsive

vlastv:
Why is `VOLUME /var/solr` instruction omitted everywhere?

dsmiley:
Good catch @vlastv !  It had been omitted deliberately because formerly the locations of variable/changing data were intermixed with some static files that shipped in the image (the configSets) which were of non-trivial size.  But now that it's separated in 8.0, it's not an issue anymore!

vlastv:
Please, update test images.

makuk66:
- `dockersolr/docker-solr:8.0.0-e5b541096f5d0a93f984311a34301bc0d1acf75b`
- `dockersolr/docker-solr:8.0.0-slim-e5b541096f5d0a93f984311a34301bc0d1acf75b`
- `dockersolr/docker-solr:8.0.0-alpine-e5b541096f5d0a93f984311a34301bc0d1acf75b`

makuk66:
I think that should be about it. Please let me know if you see any outstanding issues, otherwise we can get this shipped.

aadrian:
@makuk66 the PR looks good.

Issue #213, opened 2019-05-23
Update openjdk images


makuk66:
Per PR https://github.com/docker-library/openjdk/pull/322#issue-281393009 (and the discussion on issue https://github.com/docker-library/openjdk/issues/320), the Docker openjdk tags are changing (to https://github.com/docker-library-bot/official-images/blob/c3c6c708e231a515ae9ebb086c30ac072c7844f6/library/openjdk). Once that hits, we'll need to make some changes.
Currently, we decide what images to inherit from here: https://github.com/docker-solr/docker-solr/blob/6f7a6e812247d00f9f3c293993e26d6d041c119e/tools/update.sh#L74
So by my reading, we'll lose `openjdk:8-jre-alpine` and thus our `solr:8-alpine`s, we'll need to drop the `-jre` on the others, and we'll (automatically) drop architectures to just `amd64` for `8` and `amd64`+`arm64v8` for beyond.

henrik242:
@makuk66 Why don't you use the adoptopenjdk alpine images instead? https://hub.docker.com/r/adoptopenjdk/openjdk8

makuk66:
Because (from https://github.com/docker-library/official-images#repeatability):

> No official images can be derived from, or depend on, non-official images ...

csandanov:
I used alpine variants for my downstream solr :( sad to hear there's no more alpine-based image, and it's not just about the size, I just like having a standard base OS in all images, after all, alpine is a de-facto standard for docker images.

I've forked this repo and recovered alpine-based variants, rebased them to `adoptopenjdk`, I will try to keep it in sync with this one. Here you go https://github.com/wodby/base-solr

docwhat:
Deprecated or unused tags should be deleted to prevent people from using the obsolete and insecure images.

makuk66:
> Deprecated or unused tags should be deleted to prevent people from using the obsolete and insecure images.

That comment seems to have little or nothing to do with this (closed) issue; please use new issues. And I disagree: if someone is using an old image, or some infrastructure pulls/runs old images, it's not up to us to break them. And updating to newer versions should involve some investigation of release notes rather than a blind update. https://hub.docker.com/_/solr/ lists the current versions, so you can easily see what newer versions may be available.

PR #215, opened 2019-05-25
Remove openjdk alpine and jre

makuk66:
Fixes https://github.com/docker-solr/docker-solr/issues/213


makuk66:
Replaced by https://github.com/docker-solr/docker-solr/pull/216

Issue #217, opened 2019-06-04
OOM Handler when solr runs in foreground


das-peter:
Currently there's a potential issue when solr runs in the foreground regarding the OOM (java OutOfMemoryError) handling: https://issues.apache.org/jira/browse/SOLR-9177

The default setting of this image is to run solr in the foreground (which allows access to the console log using the `docker logs` command.). 
However, this seems to imply that java.lang.OutOfMemoryError errors won't be handled by the default `/opt/solr/bin/oom_solr.sh`, possibly leading to an unresponsive solr instance.

I've been playing around with the `-a` flag (`$SOLR_ADDL_ARGS`) when starting solr. Unfortunately I wasn't able to get it working with the default value from the [start script](https://github.com/apache/lucene-solr/blob/master/solr/bin/solr#L2123): `-XX:OnOutOfMemoryError=$SOLR_TIP/bin/oom_solr.sh $SOLR_PORT $SOLR_LOGS_DIR`
The spaces seem to cause trouble when using something like this `solr -f -a "-XX:OnOutOfMemoryError=/opt/solr/bin/oom_solr.sh 8983 /opt/solr/server/logs"`
in the default start file [`scripts/solr-foreground`](https://github.com/docker-solr/docker-solr/blob/master/scripts/solr-foreground)

So I've put the command `$SOLR_TIP/bin/oom_solr.sh $SOLR_PORT $SOLR_LOGS_DIR` into a custom script and simply used that as value for `-XX:OnOutOfMemoryError=`
That way the server seems to start just fine. However, I still have to test if the OOM handling actually kicks in and if I actually can use the env variables in the extra script or if they won't be available.

I guess for now this is just a reference to a potential problem in the hope that if someone runs into the same issue there's something to google for ;)

das-peter:
I was able to trigger a OOM error by reducing the available memory using `SOLR_JAVA_MEM: "-Xms25m -Xmx25m"`.
1. The good news is the OOM script is properly triggered.
2. The bad news is that the ENV variables are not available. So it looks like it would be quite hard to solve this in a generic way - at least in the versions before 8. As of 8, SOLR_PORT / SOLR_HOME are defined in `/etc/default/solr` which we probably can use to solve this in a generic way.
3. The other bad news is that the default script doesn't seem to actually do anything. It creates the log file which contains `Running OOM killer script for process 1 for Solr on port 8983
[...]

PR #219, opened 2019-06-07
Make OOM (OutOfMemory) handling possible

das-peter:
Pull-Request related to #217


6.6/Dockerfile
```diff
@@ -76,11 +76,19 @@ RUN set -e; \
   chown -R "$SOLR_USER:$SOLR_GROUP" /opt/solr /docker-entrypoint-initdb.d /opt/docker-solr && \
   chown -R "$SOLR_USER:$SOLR_GROUP" /opt/mysolrhome
 
+ENV TINI_VERSION v0.18.0
+ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini-static /tini-static
```
makuk66: Follow https://github.com/docker-library/official-images/blob/master/README.md#init and put it in /usr/local/bin

6.6/scripts/oom_solr.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+# This is just a wrapper to make it actually work in the start script - because
+# the whitespaces of belows call cause havoc otherwise:
+#. /opt/solr/bin/oom_solr.sh ${SOLR_PORT:-8983} /opt/solr/server/logs
```
makuk66: I'm not entirely sure what you mean there?
I don't think you want the leading period.

6.6/scripts/oom_solr.sh
```diff
@@ -0,0 +1,19 @@
...
+
+SOLR_LOGS_DIR=/opt/solr/server/logs
```
makuk66: No need to set that; it's already set in the Dockerfiles

6.6/scripts/oom_solr.sh
```diff
@@ -0,0 +1,19 @@
...
+SOLR_PID=`ps auxww | grep start.jar | grep -v grep | awk '{print $2}'`
```
makuk66: Use `$()` instead of backticks for greater readability and consistency with L14

6.6/scripts/oom_solr.sh
```diff
@@ -0,0 +1,19 @@
...
+if [[ -z "$SOLR_PID" ]]; then
+  echo "Couldn't find Solr process running!"
+  exit
+fi
+
+NOW=$(date +"%F_%H_%M_%S")
+(
+echo "Running OOM killer script for process $SOLR_PID for Solr"
+kill -9 $SOLR_PID
```
makuk66: That environment variable wants double quoting

6.6/scripts/oom_solr.sh
```diff
@@ -0,0 +1,19 @@
...
+echo "Killed process $SOLR_PID"
+) | tee $SOLR_LOGS_DIR/solr_oom_killer-$SOLR_PORT-$NOW.log
```
makuk66: That argument too

6.6/scripts/solr-foreground
```diff
@@ -16,4 +16,4 @@ fi
 
 echo "Starting Solr $SOLR_VERSION"
 
-exec solr -f "$@"
+exec solr -f -a "-XX:OnOutOfMemoryError=/opt/docker-solr/scripts/oom_solr.sh" "$@"
```
makuk66: Yea, hmm, that's not ideal. Let me think about that a bit more.

6.6/Dockerfile
```diff
@@ -76,11 +76,19 @@ RUN set -e; \
   chown -R "$SOLR_USER:$SOLR_GROUP" /opt/solr /docker-entrypoint-initdb.d /opt/docker-solr && \
   chown -R "$SOLR_USER:$SOLR_GROUP" /opt/mysolrhome
```
makuk66: Don't edit the individual Dockerfiles. Modify the `Dockerfile*template` files, then run `tools/update.sh` to generate all the Dockerfiles.

6.6/scripts/oom_solr.sh
das-peter: That's only the case for versions >= 8.
The only other variable available that resembles that for versions <8 is the optional `MY_SOLR_LOGS_DIR`.
I've changed the >= 8 script to not re-define `SOLR_LOGS_DIR` and the <8 script to use `MY_SOLR_LOGS_DIR` if available and fallback to `/opt/solr/server/logs` otherwise.

6.6/scripts/oom_solr.sh
das-peter: When testing I wanted to be as close as possible to the original handler: https://github.com/apache/lucene-solr/blob/master/solr/bin/oom_solr.sh
However, I've incorporated all your feedback into the latest update.

6.6/scripts/oom_solr.sh
das-peter: Originally I just wanted to trigger the oom_solr.sh shipped by solr https://github.com/apache/lucene-solr/blob/master/solr/bin/oom_solr.sh
But it seems to be reasonable to use our own implementation by now.

6.6/Dockerfile
das-peter: Done. But I've adjusted the command to fit better in our existing gpg handling. Hope that is ok.

6.6/Dockerfile
das-peter: Done.

6.6/scripts/solr-foreground
das-peter: Yeah, this is a nasty one.
[`$SOLR_ADDL_ARGS`](https://github.com/apache/lucene-solr/blob/master/solr/bin/solr#L2124) seems to be exactly there for stuff like this but I couldn't get it working. I always struggled with the escaping of the path to the `oom_solr.sh` script. 
If we manage to use `$SOLR_ADDL_ARGS` we probably could transparently adjust it to contain the OOM handling if not defined.

das-peter:
I've tried to address all feedback above.
Tested the image build and run of following versions 6.6, 6.6-slim & 8.1

makuk66:
Addressed by https://github.com/docker-solr/docker-solr/commit/5c9a22c4ed7b7b08fdd11b3a858001d225d2166c

Issue #222, opened 2019-06-21
Building images of the Prometheus Exporter


HoustonPutman:
Currently there is no official docker image for the [Solr Prometheus Exporter](https://lucene.apache.org/solr/guide/8_1/monitoring-solr-with-prometheus-and-grafana.html).

Prometheus is a common way of retrieving metrics from applications running in containers, so it would be nice to provide an easy way of running the prometheus exporter when running containerized Solr Cloud.

Ideally a Prometheus Exporter container would be built alongside all versions of Solr for which it exists.

makuk66:
Per https://lucene.apache.org/solr/guide/8_1/monitoring-solr-with-prometheus-and-grafana.html:

> The Prometheus exporter is included in Solr as a contrib, and is located in contrib/prometheus-exporter in your Solr instance.

and indeed I see:

```
solr@7e10b73ad9f3:/opt/solr-8.1.1$ find . -name '*prome*'
./contrib/prometheus-exporter
./dist/solr-prometheus-exporter-8.1.1.jar
```

Looking at https://lucene.apache.org/solr/guide/8_1/monitoring-solr-with-prometheus-and-grafana.html#starting-the-exporter it looks like you just run it, with appropriate configuration, which presumably you could do in a container.

So what is it you're proposing "would be built"?

HoustonPutman:
Oh I guess I didn't look around in the image fully enough. Thanks for the help, I'll test that out.

dsmiley:
It would be helpful to tell users how to run the prometheus exporter from Docker; it's not obvious!
I got this to work... obviously needs adjustments based on your deployed environment.
````
docker run --rm --name solr-exporter -p 9983:9983 \
  --entrypoint=sh \
  --workdir=/opt/solr/contrib/prometheus-exporter/ \
  solr:8 \
   bin/solr-exporter -b http://host.docker.internal:8983/solr
````
The fact that bin/solr-exporter isn't executable is a small problem we should fix.  If it's executable upstream, it ought to be executable likewise in the image.

PR #224, opened 2019-06-26
Adding prometheus exporter support

HoustonPutman:
Relates to issue #222 

There was an issue with the prometheus-executor binary not having executable permissions in the 8.x docker images, this is not an issue for the non-`varsolr` docker template.

I also created a startup script for the exporter, that only gets added for Solr versions where the prometheus exporter exists (>= 7.3).


scripts-prometheus-exporter/prometheus-exporter
```diff
@@ -0,0 +1,43 @@
...
+
+if [[ -z "$EXPORTER_CONFIG" ]]; then
+    EXPORTER_CONFIG=/opt/solr/contrib/prometheus-exporter/conf/solr-exporter-config.xml
+fi
+
+if [[ ! -z "$EXPORTER_CONFIG" ]]; then
+    ARGS="$ARGS -f $EXPORTER_CONFIG"
+fi
+
+if [[ ! -z "$EXPORTER_PORT" ]]; then
```
makuk66: shellcheck likes to see -n instead of ! -z

scripts-prometheus-exporter/prometheus-exporter
```diff
@@ -0,0 +1,43 @@
...
+    set -x
+fi
+
+ARGS=
+
+if [[ -z "$EXPORTER_CONFIG" ]]; then
+    EXPORTER_CONFIG=/opt/solr/contrib/prometheus-exporter/conf/solr-exporter-config.xml
+fi
+
+if [[ ! -z "$EXPORTER_CONFIG" ]]; then
```
makuk66: You've just set that, so the test is not needed.

scripts-prometheus-exporter/prometheus-exporter
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
```
makuk66: Hm, I don't really want a proliferation of random script directories at the top level; I was already a bit unhappy with having both `scripts` and `scripts-before8`. It would be better to just add `prometheus-exporter` to the normal scripts directories, and then in `update.sh` (inverting the logic you added) remove it from old version directories. Or you could simply just add it everywhere, and then in the script itself test with `-f` if `Solr-exporter` is available in the filesystem, and error out with "Prometheus is not available prior to Solr 7.3" if not; then you don't need any `build.sh`/`update.sh` changes, which will keep things easier to maintain. Do that.

Looking at the contents, it seems to me you're just translating environment variables to command-line arguments (rather than say use determining some run-time information), so they need to be configured outside of the container and get passed in. At which point, what's the value of this over just providing those arguments to the command to run form your calling environment? Out of interest, what is your orchestration environment (I know OpenShift makes doing that inconvenient)?

tools/build.sh
```diff
@@ -34,12 +34,19 @@ fi
 
 echo "Updating scripts"
```
makuk66: Great, you found the right spots to do this :-) At some point we may want to extract this into a common place so both `update.sh` and `build.sh` can call the same bit of code.

scripts-prometheus-exporter/prometheus-exporter
HoustonPutman: I like that a lot more. (Checking for the executable in the script)

Our orchestration framework is Kubernetes, and we are looking to add metrics capabilities in the [solr-operator](https://github.com/bloomberg/solr-operator). 

I think there are 2 good reasons to have an interface (the `prometheus-exporter` script) that obfuscates the base executable.
- It allows for the user to be unaware of any changes solr makes to the location of the executable, new start scripts, or small changes in the command line parameters.
- It also makes it possible to have custom builds of solr docker imitate this interface. That way users can drop in their custom images, and infrastructure that uses the official docker image will be able to interact with the custom image in the same way.

HoustonPutman: Thanks for such a thorough review! I think I've cleaned everything up.

I'd understand if you prefer to leave out the `prometheus-exporter` script, as the only required change is adding `chmod ..../solr-exporter` in `Dockerfile-varsolr.template`. I just think the script adds an improved user experience :)

makuk66:
I had another pass through this. There are still issues with the argument passing. Have a look at https://github.com/docker-solr/docker-solr/pull/233 -- note the changed quoting in eg `ARGS+=(-p "$EXPORTER_PORT")`, and the `"${@}"` expansion in the exec. Now if you test with arguments containing spaces the arguments are passed through properly.
I also added `set -u` and cleared up the resulting spellcheck warnings. I've removed the executability checking, because I'll add permission setting to the Dockerfiles; and if you override it with a mounted image you'll get an error that is clear enough that an extra check isn't all that useful.

The main thing that happens by default is the setting of the `-f`. According to https://lucene.apache.org/solr/guide/8_1/monitoring-solr-with-prometheus-and-grafana.html#command-line-parameters that is the default, but that is incorrect; the default is `conf/solr-exporter-config.xml`, so if you were in `contrib/prometheus-exporter` it should just work without that. I've added a change that does that `cd`.

> - It allows for the user to be unaware of any changes solr makes to the location of the executable, new start scripts, or small changes in the command line parameters.
> - It also makes it possible to have custom builds of solr docker imitate this interface. That way users can drop in their custom images, and infrastructure that uses the official docker image will be able to interact with the custom image in the same way.
 
I'm not wholly convinced. The first point sounds like a solution to a problem that doesn't exist; and if changes do happen a user should be reviewing what they are anyway. For the second point, if someone goes through the effort to provide a custom script, it's not much of an effort to modify the command-line to point to its path.

And you will need to specify a path to the script anyway, because this script is not the default in the docker entry point. And if you're going to do that, then you might as well pass in whatever arguments you need to there; at which point does script really doesn't add a lot, and really just invents another layer of abstraction to contend with.

Anyway, I am willing to add this if you feel it assists your use-case, but I can't test it easily, as I don't have a Prometheus setup at the moment. Can you test it for me? We won't get it in for 8.2, but we can for the next one (and of course you can try the script in your own images ahead of that time).

HoustonPutman:
I'm fine dropping the script if you don't believe it provides value, as it would only add more work for y'all in the future. The only thing that is absolutely necessary is the `chmod` command on the prometheus exporter directory. And I'll be able to test that for you quite easily.

makuk66:
OK, then let's leave it for the moment. If someone wants it, they can grab it from https://github.com/docker-solr/docker-solr/blob/fd5cd1450d7e7f4b449dea69274ae411d3065ed2/scripts/prometheus-exporter
If there turn out to be other needs for it, or we discover other useful functionality it can provide, add comments here or file an issue, and we can soon pop it back.

PR #233, opened 2019-07-28
Prometheus

makuk66:


makuk66:
See https://github.com/docker-solr/docker-solr/pull/224#issuecomment-516418303

Issue #261, opened 2019-11-21
Use closest Apache mirror as default download source


janhoy:
Today we default to archive.apache.org for pulling Solr tarball, with the option of specifying a `SOLR_DOWNLOAD_SERVER` var.

However the default should be to resolve closest mirror using `closer.lua` and if package not found there, fallback to `www.apache.org/dist` and if not found there fallback to archive.

See https://hub.docker.com/r/harisekhon/hbase/dockerfile for an example from Hbase.

makuk66:
> Today we default to archive.apache.org for pulling Solr tarball, with the option of specifying a SOLR_DOWNLOAD_SERVER var.

That choice was made because we hit various transient and non-transient failures with mirrors, with some mirrors missing new releases that had not yet arrived, or older releases that they chose not to host. This caused lots of failed builds, wasted time, frustration and delayed releases. Archive consistently worked, so that's what I used. Because it is mainly used by docker hub at build time, not by users at run time, it should not provide a heavy load on archive, though I have no metrics to know.

I can't escape the feeling that this day and age you should just use a CDN, host all versions, have a single URL for each version that is served by the CDN's local edge location, rather than force users having to deal with this.

> However the default should be to resolve closest mirror using closer.lua and if package not found there, fallback to www.apache.org/dist and if not found there fallback to archive.

For backwards compatibility the SOLR_DOWNLOAD_SERVER should continue to work if set.

I may have some time to have a shot at this.

PR #262, opened 2019-11-21
Use the closest Apache mirror

makuk66:
Use the closest Apache mirror

Fixes https://github.com/docker-solr/docker-solr/issues/261


janhoy:
This looks ok. A benefit besides playing nice with Apache owned bandwidth is that users building the image will get faster build times with a closer mirror. This is the CDN of the 90's :-) and univerities and others donate their own storage and bandwidth to the ASF for free thus indirectly sponsoring the CDN.

Fun fact: Apache OpenOffice has so massive download numbers that they continue to host their artifacts with SourceForge since they were afraid to overload ASF servers and mirror network :-)

makuk66:
Merged https://github.com/docker-solr/docker-solr/commit/70384cb4394d1bcc82f53dd14f2593f4b574df3f

PR #268, opened 2019-12-06
Add vagrantfile

makuk66:


vagrant/README.md
```diff
@@ -0,0 +1,12 @@
+This directory contains a [Vagrantfile](Vagrantfile) for building docker-solr.
+
+Pre-requisites: [VirtualBox](https://www.virtualbox.org/wiki/Downloads) and [Vagrant](https://www.vagrantup.com/downloads.html).
+
+```bash
+vagrant up
+vagrant ssh -c "git clone git@github.com:docker-solr/docker-solr.git && \
+  cd docker-solr && \
+  ./tools/update.sh &
```
janhoy: Typo? Should be && \

vagrant/README.md
```diff
@@ -0,0 +1,12 @@
+This directory contains a [Vagrantfile](Vagrantfile) for building docker-solr.
+
+Pre-requisites: [VirtualBox](https://www.virtualbox.org/wiki/Downloads) and [Vagrant](https://www.vagrantup.com/downloads.html).
```
janhoy: Note: Latest version of vagrant does not support latest version of virtualbox, i.e. vagrant 2.2.6 needs virtualbox 6.0

vagrant/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+```bash
+vagrant up
+vagrant ssh -c "git clone git@github.com:docker-solr/docker-solr.git && \
```
janhoy: Could not get the git clone to work with ssh, changed to https://github.com/docker-solr/docker-solr.git and it worked

vagrant/README.md
makuk66: Thanks, Fixed

vagrant/README.md
makuk66: Workarounds are described on [Oracle's blog](https://blogs.oracle.com/scoter/getting-vagrant-226-working-with-virtualbox-61-ga). Others are running into this issue too, see eg [this discussion](https://github.com/oracle/vagrant-boxes/issues/178). It has been filed upstream on https://github.com/hashicorp/vagrant/issues/11249, where it will be fixed in the next release. The maintainer [commented here](https://github.com/hashicorp/vagrant/issues/11249#issuecomment-565214999) which I interpret to be soon'ish. So I don't think we need to specifically add this to the README.

vagrant/README.md
makuk66: I personally prefer the [ssh URLs](https://help.github.com/en/github/authenticating-to-github/connecting-to-github-with-ssh) for repos I have write access to, and I have my git configured such that this works. But, [github recommends https](https://help.github.com/en/github/using-git/which-remote-url-should-i-use), afaict mainly because it works more widely. I'm happy to change it to that.

janhoy: Yes, unless we document prerequisites for the ssh method to work. Are we going to push changes to the repo from within vagrant?

makuk66: I won't be using vagrant myself, but for those who do, presumably yes. If you git push that with a https URL you get prompted for credentials, so that will work too.

janhoy:
I am experimenting with a top-level `Vagrantfile` which mounts up CWD as `/vagrant` in the vm, and thus updates the host file system instead of the guest file system. Doing it this way you can use `vagrant ssh ...` in-place for commands you don't have locally on your OS.

janhoy:
> I am experimenting with a top-level `Vagrantfile`

So after #271 I have abandoned the idea of a top-level Vagrantfile. It could have been a good idea on macOS, but for Windows, it makes more sense to run all commands inside the VM. I'm not on Windows so I won't have time to pursue this.

I encourage you to merge #271 first and then add the docs from this PR into a new "Windows" section in update.md. Then merge this PR with a big disclaimer that it needs more testing.

janhoy:
I added a proposed fix to update.md to resolve the conflict

vagrant/Vagrantfile
```diff
@@ -0,0 +1,16 @@
+Vagrant.configure("2") do |config|
+  config.vm.box = "ubuntu/bionic64"
+  config.vm.provider "virtualbox" do |v|
+    v.memory = 2048
+    v.cpus = 2
+  end
+ config.vm.provision "shell", inline: <<-SHELL
+   apt-get update
+   apt-get -y install lsof procps curl wget gpg gawk shellcheck vim less git
```
janhoy: Need to add parallel as new dependency here

PR #272, opened 2020-01-05
Support for SOLR_DOWNLOAD_URL and other build-args

### LINKED ISSUE #264: Build-arg for custom Solr download URL
Sometimes solr users need to fix and roll out bugs and roll out a patched version in their environment. Today this requires quite some changes to `Dockerfile` in a downstream fork.

I propose to move a few `ENV`s over to `ARG` with default values:

    ARG SOLR_VERSION="8.3.1"
    ARG SOLR_SHA512="75472c1....fff"
    ARG SOLR_DOWNLOAD_URL

The first two simply replace the existing vars but lets you override them. Then you can replace the literal string `8.3.1` with `$SOLR_VERSION` elsewhere in Dockerfile.

The new arg `SOLR_DOWNLOAD_URL` will not be used if not specified. If specified it will work similar to `SOLR_DOWNLOAD_SERVER` but allow for a custom URL outside of the mirrors. When this is used you need to supply your own `SOLR_SHA512` arg for download validation, and Dockerfile will then skip GPG validation.

I'll issue a PR for this soon :)

janhoy:
The goal of this PR is to make the Dockerfiles from this project easier to use by developers, and at the same time make it more DRY. Here's a list of improvements

* Converted ENV `SOLR_VERSION` and `SOLR_SHA512` into build-args
* Using SHA512 instead of SHA256 since this is what upstream already uses
* Added new build-arg `SOLR_DOWNLOAD_URL` (#264)
  This enables downstream users to build a docker image with a custom build Solr version by also passing in the build-arg SOLR_SHA512
  In case of custom URL, we skip GPG validation as this tgz will not be signed
* We clear a bunch of env vars that are only cluttering the image runtime. This is done in `docker-entrypoint.sh` since there is no UNSETENV command

Fixes #264

### DIFF
--- a/Dockerfile-varsolr.template
+++ b/Dockerfile-varsolr.template
@@ -127,10 +127,7 @@ RUN set -ex; \
   chown -R "0:0" /opt/solr-$SOLR_VERSION /docker-entrypoint-initdb.d /opt/docker-solr; \
   chown -R "$SOLR_USER:0" /var/solr; \
   { command -v gpgconf; gpgconf --kill all || :; }; \
-  rm -r "$GNUPGHOME"; \
-  # Clear some variables that are of no use runtime
-  unset GOSU_KEY GOSU_VERSION TINI_KEY TINI_VERSION SOLR_USER SOLR_UID SOLR_GROUP SOLR_GID \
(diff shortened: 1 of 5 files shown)


janhoy:
Added `SOLR_KEYS` as build arg. See #274 for an alternative way where we don't need per-release configuration of keys.

Issue #278, opened 2020-01-14
Update Solr 8.4.0 to 8.4.1


janhoy:
@dsmiley do you want to give it a shot and test the deploy stuff on mac?

dsmiley:
I will @janhoy .

dsmiley:
FYI we need to remove xml-apis jar from Tika contrib -- see  https://issues.apache.org/jira/browse/SOLR-14054?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&focusedCommentId=17001228#comment-17001228
That has been on my mind and I've been meaning to file a PR.

dsmiley:
That Tika thing is actually for 8.5 so it's a non-issue right now.

I've been following instructions and pushed a PR:
#279 
I've also been watching Martijn's video screen share of the release process:
https://www.youtube.com/watch?v=K0a5lhl28Vg

It's a shame that each build, which builds all versions has to actually download a full Solr .tgz, each of which is around 180MB, sometimes less.  The existence of a "/downloads" with the distributions there doesn't help.  The root cause could be argued this is Solr's problem for being so heavyweight.  Plugin-ification will help.

makuk66:
> It's a shame that each build, which builds all versions has to actually download a full Solr .tgz

There are two downloads:

- the one done by update.sh, which is a one-off, because the files are retained in the downloads folder, where it gets found next time.
- the one done by the Dockerfiles. That one could be resolved with something like https://github.com/docker-solr/docker-solr/issues/140 further discussed in https://github.com/docker-solr/docker-solr/issues/196#issue-392195987 point 7. That was some time ago, so perhaps things have changed. I haven't explicitly discussed this pain point with the library team. The parallel building by Travis, and by Jan's changes, minimise this cost somewhat, but it remains a waste of time, especially on slow links.

 > The existence of a "/downloads" with the distributions there doesn't help. 

I'm not sure what you mean here?

> The root cause could be argued this is Solr's problem for being so heavyweight. Plugin-ification will help.

Well there is always that :-) Improvements there would be appreciated. But I still think there should be a better docker-level option here.

> I've been following instructions and pushed a PR: https://github.com/docker-solr/docker-solr/pull/279

Great!

[...]

PR #280, opened 2020-01-18
Fix generate-stackbrew-library.sh to work on macos

janhoy:
Fixed for macos. Depends on GNU version of `find` so had to install that as well :) 

This should help with #278 on Macos

### DIFF
--- a/tools/init_macos.sh
+++ b/tools/init_macos.sh
@@ -1,0 +1,12 @@
+# Source this file to prepare a mac for running scripts
+hash greadlink 2>/dev/null || { echo >&2 "Required tool greadlink is not installed, please run 'brew install coreutils'"; exit 1; }
+hash gfind 2>/dev/null || { echo >&2 "Required tool gfind is not installed, please run 'brew install findutils'"; exit 1; }
+if [[ ! -d /tmp/docker-solr-bin ]]; then
+  mkdir -p /tmp/docker-solr-bin >/dev/null 2>&1
+  ln -s /usr/local/bin/greadlink /tmp/docker-solr-bin/readlink
+  ln -s /usr/local/bin/gfind /tmp/docker-solr-bin/find
+fi
+if [[ ! $PATH == *"docker-solr-bin"* ]]; then
+  export PATH=/tmp/docker-solr-bin:$PATH
+  echo "Configuring for macOS - ádding /tmp/docker-solr-bin first in path"
+fi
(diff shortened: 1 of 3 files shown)


update.md
```diff
@@ -99,15 +99,16 @@ Using [Homebrew](https://brew.sh/), install the necessary dependencies for macOS
...
-# Make gnu readlink the default. You may wish to undo this after building 
+# Make gnu readlink and find the default. You may wish to undo this after building 
 ln -s /usr/local/bin/greadlink /usr/local/bin/readlink
+ln -s /usr/local/bin/gfind /usr/local/bin/find
 ```
 
-NOTE: If you don't want to symlink readlink permanently you may instead
-create a symbolic link to `/path/to/some/bin/readlink` and put that location
+NOTE: If you don't want to symlink readlink and gfind permanently you may instead
+create symbolic links to `/path/to/some/bin/readlink` and `/path/to/somr/bin/find` and put that location
```
makuk66: `some`

makuk66: `some`

dsmiley:
Instead of symlink-ing in "find", can/should the bash script refer to "gfind" instead?  I modified it as so and it worked.  Perhaps "gfind" isn't referred to as such on a linux host but only in Mac OS with brew?  I don't know.

janhoy:
We could have a check at the start of each main script to check for correct versions of tools and tell the user to run `. macos_init.sh` which would do the required setup for this shell only.

dsmiley:
That works for me... it's a shame there are these troublesome platform differences.  I like this "this shell only" aspect here; I don't want to symlink core executable utilities in my system if I can help it.

janhoy:
The last commit adds an init script that can be sourced to put a folder with symlinked GNU tools first in path. User needs to source this in each shell before starting to run commands...

tools/init_macos.sh
```diff
@@ -0,0 +1,12 @@
...
+hash greadlink 2>/dev/null || { echo >&2 "Required tool greadlink is not installed, please run 'brew install coreutils'"; exit 1; }
+hash gfind 2>/dev/null || { echo >&2 "Required tool gfind is not installed, please run 'brew install findutils'"; exit 1; }
+if [[ ! -d /tmp/docker-solr-bin ]]; then
+  mkdir -p /tmp/docker-solr-bin >/dev/null 2>&1
+  ln -s /usr/local/bin/greadlink /tmp/docker-solr-bin/readlink
+  ln -s /usr/local/bin/gfind /tmp/docker-solr-bin/find
+fi
+if [[ ! $PATH == *"docker-solr-bin"* ]]; then
+  export PATH=/tmp/docker-solr-bin:$PATH
+  echo "Configuring for macOS - ádding /tmp/docker-solr-bin first in path"
```
dsmiley: Why the accent on the 'a' in "adding"?

janhoy: Oops :)

Issue #283, opened 2020-01-20
Use JRE instead of JDK and publish only 'slim' variant


janhoy:
Our default `docker run solr` image is 824Mb uncompressed. If we always use 'slim' variant (why would we need full?) and also use JRE intead of JDK (why would we need jdk?), then we'd reduce that default image experience to about 433Mb.

We can remove the "fat" debian by simply tagging all versions both as x.y.z and x.y.z-slim, i.e. 8.4.1 would be tagged `8.4.1, 8.4.1-slim, 8.4, 8.4-slim, 8, 8-slim, latest, latest-slim` and we can remove half our repo files and half the build/test time :)

Mandalka:
Not inspected deeper and sorry f not relevant (until now very few small packages seem to be installed by apt) / yet known.

Depending on base image settings and metadata of the additional installed packages using apt parameter no-install-recommends can save space, too (but sometimes more for debugging to add further needed packages which else are installed by recommeds) by
apt-get install --no-install-recommends ...

makuk66:
This could impact folks who have custom Dockerfiles that use ours as `FROM` and depend on the presence of things that are not in `slim`. That's the sort of possibly backwards-incompatible change I would usually leave to the next major Solr version. But I do think the risk is probably low enough that I'm not too worried if that's what you want to do. The change looks good

janhoy:
I can be talked into doing this from 8.5 and onwards only, but that means more build complexity, a new template file etc. Do you have a way to "ask" the community/users?

dsmiley:
This seems like something to put off until 9.0.

makuk66:
> Do you have a way to "ask" the community/users?

Only these GitHub issues

> This seems like something to put off until 9.0.

That's my feeling too.

ctargett:
#284 is merged, so this seems like the right place to bring this up, but I think we need to reconsider the removal of the full JDK. Moving to the JRE removed several tools like `jcmd` and `jmap` which are essential tools when debugging heap usage in a production cluster. This is an incredibly common need when looking at performance issues like excessive memory use, and it's from this level of analysis that at least a few memory leaks in Solr have been discovered.

What alternatives are recommended for this sort of profiling instead?

janhoy:
This sounds like a valid argument for sticking with JDK for sure. Should be perhaps provide a separate `-jdk` tag option so people can opt in? I suppose there is no way to just install these tools without the full JDK?

dsmiley:
[...]

PR #284, opened 2020-01-20
Use JRE instead of JDK as standard, to reduce image size

janhoy:
Fixes first part of #283, i.e. using only JRE base image.

I'll open another PR for only publishing Debian 'slim' variant.


janhoy:
I tested building and spinning up an image. Think it should be safe. Martijn, any reason to stick with jdk?

makuk66:
There was a reason, but I'm struggling to recall it. Perhaps from [Apache Solr 3 Enterprise Search Server](https://books.google.co.uk/books?id=ChKVwotW8mYC&pg=PT58&lpg=PT58&dq=solr+jre+jdk&source=bl&ots=UYbCXnN3hO&sig=ACfU3U1lXtH-8bfj7CE3Iqq9wGNL3jWlYw&hl=en&sa=X&ved=2ahUKEwiEw8KT2pXnAhU0olwKHcwPC1Y4FBDoATAHegQIChAB#v=onepage&q=solr%20jre%20jdk&f=false) "so the JDK is recommended" with reference to `jconsole` and other diagnostics tools? Or are there any plugins or contributors modules that need it? I expect that the JRE is just fine for the vast majority.

janhoy:
> There was a reason, but I'm struggling to recall it.

Before Solr 4 (?) we had JSP Admin UI which required the compiler to compile the JSP pages. JConsole is no longer part of the jdk. Guess people can make their own image if they need more software in it.

PR #288, opened 2020-01-21
Only publish slim images

janhoy:
Part of #283. This PR depends on and includes the changes from #284 so should be merged after that one.


janhoy:
Brought the PR up to date with master and the other changes.
I think the most important is back-compat for people who specify `solr:slim` or `solr:8-slim`, they will still get what they want. Folks who use our image in FROM must of course prepare to apt-get more stuff in their downstream image, but it will break in their build/CI, not in production.

janhoy:
I'm going to push a change to this PR that will continue building dual images for all historic versions, but from v9.0 they will be slim only (without any slim suffix or sub folder since it is default). That way we don't mess with back compat, and we get time to alert people about the change. 

Since this will interact with #289, I propose that we fold that change into this PR as well (if there is consensus for using AdoptOpenJDK).

janhoy:
As 9.0 is approaching, I think this work will be replaced with changes to the official Dockerfile in "solr" repo, so that will be slim only, and perhaps use Temurin JDK?

Also, `docker scan` is a useful tool to discover garbage in our images that can be removed. I.e. we don't need to ship with gpg and dependencies, and we don't need to ship with perl, both of which cause several scan warnings.

janhoy:
In the main project we already only publish slim image. Closing.

PR #293, opened 2020-02-27
Fix broken README link to init-var-solr

uosl:
The file had been renamed from init-var-solr.sh to init-var-solr, but the README not updated.


dsmiley:
Thanks!

PR #218, opened 2019-06-06
Solr 7.7.2

makuk66:

PR #295, opened 2020-03-24
Added newly released Solr 8.5

jaceq:
As per title.
I also switched to debian buster for regular build and given than slim used latest I explicitly defined buster as base image.
Image builds and solr starts.

### DIFF
--- a/8.5/Dockerfile
+++ b/8.5/Dockerfile
@@ -1,0 +1,140 @@
+
+FROM openjdk:11-jre-buster
+
+LABEL maintainer="Martijn Koster \"mak-docker@greenhills.co.uk\""
+LABEL repository="https://github.com/docker-solr/docker-solr"
+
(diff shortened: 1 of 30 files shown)


makuk66:
It looks like you manually copied the files and made changes? That's not how this works. For instructions see [update.md](https://github.com/docker-solr/docker-solr/blob/master/update.md), and note that it also updates other files (see [example](https://github.com/docker-solr/docker-solr/pull/218/files#diff-4819dc9f2283ad2dac18ec754438428b)). If you want to update the FROM (which I think is more appropriate for a major version), then that involves changing the [script](https://github.com/docker-solr/docker-solr/blob/master/tools/update.sh#L79) that generates the docker files.

jaceq:
You I right, I copied this. 
I will close this PR and try the 'supported' way.
Thanks for letting me know.

PR #296, opened 2020-03-25
Added solr 8.5 / 8.5.0

jaceq:
This is a re-try of https://github.com/docker-solr/docker-solr/pull/295 

I did as advised by @makuk66 , used scripts to generate 8.5 tree, did separate commit to add new key, run build and tests scripts.

This time I have not changed the FROM as before. 
Please review.


janhoy:
Thanks for contributing. This looks right. I suppose you got the warning that -stretch is "not supported". But I checked the openjdk images, and both -stretch and -buster have the same Java versions, so we should be safe sticking to -stretch for 8.5 if we want. Then we can go to buster from 9.x.

I'll merge your PR after some more checking and then file a PR to get it to docker-hub

Issue #310, opened 2020-05-07
Script wait-for-solr.sh is not working


jeetchris:
Hi,

I have been trying to use the wait-for-solr.sh script to initialize a solr instance in local (Mac, docker-machine).
But the script is simply replying that solr is not running yet.
I guess the issue is related to bash options used at the top of the script, because if I run the wget command directly inside the container, it works correctly.
Solr Admin UI is also accessible from Mac.

I can reproduce the issue on both solr:7.1.0 and solr:8.5.1 images.

Also, if I use the script directly within solr machine (so using localhost by default), it just hangs. I suppose until timeout, but did not wait a long time.

docker-compose.yml
```yaml
version: '3'
services:
  solr:
    image: solr:8.5.1
    ports:
      - 8983:8983
  solr-init:
    image: solr:8.5.1
    command: bash -c "wait-for-solr.sh --solr-url http://solr:8983"
```
```bash
$ docker-compose up
Creating network "solr_default" with the default driver
Creating solr_solr_1      ... done
Creating solr_solr-init_1 ... done
Attaching to solr_solr-init_1, solr_solr_1
solr_1       | Starting Solr
solr_1       | OpenJDK 64-Bit Server VM warning: Failed to reserve shared memory. (error = 1)
solr_1       | OpenJDK 64-Bit Server VM warning: Failed to reserve shared memory. (error = 1)
solr_1       | OpenJDK 64-Bit Server VM warning: Failed to reserve shared memory. (error = 1)
solr_1       | 2020-05-07 04:18:15.152 INFO  (main) [   ] o.e.j.u.log Logging initialized @1696ms to org.eclipse.jetty.util.log.Slf4jLog
solr_1       | 2020-05-07 04:18:15.506 INFO  (main) [   ] o.e.j.s.Server jetty-9.4.24.v20191120; built: 2019-11-20T21:37:49.771Z; git: 363d5f2df3a8a28de40604320230664b9c793c16; jvm 11.0.7+10
solr_1       | 2020-05-07 04:18:15.577 INFO  (main) [   ] o.e.j.d.p.ScanningAppProvider Deployment monitor [file:///opt/solr-8.5.1/server/contexts/] at interval 0
solr_1       | 2020-05-07 04:18:16.052 INFO  (main) [   ] o.e.j.w.StandardDescriptorProcessor NO JSP Support for /solr, did not find org.apache.jasper.servlet.JspServlet
[...]

PR #311, opened 2020-05-08
Fix race condition in wait-for-solr.sh

jeetchris:
Solve #310.

Script is failing to detect Solr instance as alive due to a race condition between two commands in a pipe.

> `grep -q` succeeds and exits early, causing the still-writing `wget` to fail to write its stdout to the pipe, and the overall pipeline to fail.

Solution is:

> replacing the grep part of the pipeline with just: grep -i solr >/dev/null

I tested it for Solr 8.5.1, under Mac and docker-machine only.

Issue can be reproduced (depending on the environment), using this docker-compose.yml file.

```yaml
version: '3'
services:
  # Solr instance
  solr:
    image: solr:8.5.1
    ports:
      - 8983:8983
  # Should detect Solr running successfully
  solr-success:
    image: solr:8.5.1
    command: bash -c "wait-for-solr.sh --solr-url http://solr:8981"
  # Should not detect Solr, since port is wrong
  solr-fail:
    image: solr:8.5.1
    command: bash -c "wait-for-solr.sh --solr-url http://solr:8980"
```

### DIFF
--- a/5.5/scripts/wait-for-solr.sh
+++ b/5.5/scripts/wait-for-solr.sh
@@ -79,7 +79,7 @@ grep -q -E '^https?://' <<<"$solr_url" || usage "--solr-url $solr_url: not a URL
 
 ((attempts_left=max_attempts))
 while (( attempts_left > 0 )); do
-  if wget -q -O - "$solr_url" | grep -q -i solr; then
+  if wget -q -O - "$solr_url" | grep -i solr >/dev/null; then
     break
   fi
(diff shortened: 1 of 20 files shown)


jeetchris:
@makuk66 Thank you, I ran the command and committed the result.
No new Solr version was found apparently.
Tell me if you prefer that I combine both commits into one.

makuk66:
Great stuff -- thanks!

Issue #276, opened 2020-01-07
Prepare transfer of project to the Apache Lucene project


janhoy:
See https://issues.apache.org/jira/browse/SOLR-14168 for a discussion in the Solr community about how to maintain this project going forward.

makuk66:
Just to add a bit here: my personal circumstances are changing, and I won't be able to continue putting in the volunteer hours to maintain this image. To ensure continued availability, and further releases, I have approached Solr committers @janhoy and @dsmiley, who have expressed an interest in the further development and maintenance of this project (for which my thanks) and can work towards tighter integration with the Solr project itself. To that end I have made a Software Grant Agreement to transfer the project to the ASF, and it has been adopted by the PMC.

For end-users, there will be no changes to the image. The actual images are built by the official library team, and hosted on Docker hub, as always.

There will be changes to the project to make it easier for the new maintainers to manage and integrate with the Solr project, such as changes to CI infrastructure, who does the release and when, and perhaps how bugs are handled, and what/where documentation is available. The GitHub repo itself probably won't change much, as it is the way it is because of how the docker library builds images. These things will be figured out over time, in discussions in [SOLR-14168](https://issues.apache.org/jira/browse/SOLR-14168) and on the [solr-dev mailing list](https://lucene.apache.org/solr/community.html#developer-list-devluceneapacheorg).

For the docker library team @tianon @yosifkit there will be some new faces, but I don't think there will be a lot else different.

I will continue to be participating in this repo and as advisor, for an interim period, to ensure a smooth transition.

And I'd like to take the opportunity to say a heartfelt thank-you to all the contributors who have filed issues and provided patches over the years. And to [Lucidworks](https://lucidworks.com) for the period where they supported this work. And of course to the users, who have given this image 10M+ downloads and 724 stars on Docker Hub.

I'll leave this issue open so we can add comments as more becomes known.

DwaneHall:
Thanks for all of your hard work and effort @makuk66 as a user it has been a pleasure following and using this image and your support and commitment to the project has been something to be very proud of.  Your responses to questions, example configurations, and clarity and thoughtfulness on design decisions were ever present throughout the DockerHub and GitHub repos.
 
About 6 months ago I began the journey of a large Solr infrastructure upgrade and I honestly believe the image has saved me 100's of hours of time when transitioning from a traditional bare metal Solr install.  I'm sure I'm not alone in this category.

Sincerest best wishes on your future projects both personally and professionally.

[...]

PR #313, opened 2020-05-10
Update maintainer information

janhoy:
Part of #276

### DIFF
--- a/Dockerfile-varsolr.template
+++ b/Dockerfile-varsolr.template
@@ -1,7 +1,7 @@
 
 FROM $REPLACE_FROM
 
-LABEL maintainer="Martijn Koster \"mak-docker@greenhills.co.uk\""
+LABEL maintainer="The Apache Solr/Lucene Project"
 LABEL repository="https://github.com/docker-solr/docker-solr"
 
 ARG SOLR_VERSION="$REPLACE_SOLR_VERSION"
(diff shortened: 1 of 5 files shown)


Dockerfile-varsolr.template
```diff
@@ -1,7 +1,7 @@
 
 FROM $REPLACE_FROM
 
-LABEL maintainer="Martijn Koster \"mak-docker@greenhills.co.uk\""
+LABEL maintainer="The Apache Solr/Lucene Project"
```
dsmiley, line 4: Lucene/Solr

generate-stackbrew-library.sh
```diff
@@ -69,8 +69,10 @@ getArches 'solr'
 cat <<-EOH
 # this file is generated via https://github.com/docker-solr/docker-solr/blob/$(fileCommit "$self")/$self
 
-Maintainers: Martijn Koster <mak-github@greenhills.co.uk> (@makuk66),
-             Shalin Mangar <shalin@apache.org> (@shalinmangar)
+Maintainers: The Apache Lucene/Solr Project <solr-user@lucene.apache.org>,
+             Shalin Mangar (@shalinmangar),
+             David Smiley (@dsmiley),
+             Jan Høydahl (@janhoy)
```
dsmiley, line 75: Oh come on; put yourself higher up :-)

janhoy: Strictly chronological :-)

This is just a temp change. Once we move the repo to ASF github, the maintiners list should ideally just be solr-user list? But the project does not have a github account for itself, so not sure if they will accept that part being empty?

PR #321, opened 2020-06-05
Add the jattach tool

janhoy:
As discussed in #283. See https://github.com/apangin/jattach

    Usage: jattach <pid> <cmd> [args ...]
    
      Commands: 
        load : load agent library
        properties : print system properties
        agentProperties : print agent properties
        datadump : show heap and thread summary
        threaddump : dump all stack traces (like jstack)
        dumpheap : dump heap (like jmap)
        inspectheap : heap histogram (like jmap -histo)
        setflag : modify manageable VM flag
        printflag : print VM flag
        jcmd : execute jcmd command


ctargett:
Thanks Jan, hopefully this will help!

PR #354, opened 2020-10-15
Require init_var_solr; don't setup in Dockerfile

dsmiley:
I propose removing initialization of /var/solr from the Dockerfile, thus leaving only init_var_solr to do this.  The fact that it's in two places means that the image has two solr.xml, two zoo.cfg, two log4j2.xml.  This initialization itself must be maintained twice.  That leads to confusion (it did with my colleagues and I) about which copy is going to be used.  Imagine you are basing your company Solr Dockerfile on top of this one (i.e. official is the FROM) and need to do modifications.  Do you modify /opt/solr/server/solr/solr.xml?  Surprise surprise, _sometimes_ it is copied to /var/solr/data/ by the init_var_solr script but sometimes it isn't because the Dockerfile here will do it, thus ignoring the customizations made to solr.xml in the next image layer.

After making this change, our wonderful tests exposed that solr-demo wasn't invoking init_var_solr.

Is there some reason we should _not_ do this?  There are additional lines printed:
````
Creating /var/solr/data
Creating /var/solr/logs
Copying solr.xml
Copying zoo.cfg
Copying log4j2.xml
````
Perhaps merely a single line "Initializing /var/solr" should be printed if the directory is empty?

### DIFF
--- a/8.0/Dockerfile
+++ b/8.0/Dockerfile
@@ -101,11 +101,7 @@ RUN set -ex; \
   mv /opt/solr/bin/solr.in.cmd /opt/solr/bin/solr.in.cmd.orig; \
   chown root:0 /etc/default/solr.in.sh; \
   chmod 0664 /etc/default/solr.in.sh; \
-  mkdir -p /var/solr/data /var/solr/logs; \
-  (cd /opt/solr/server/solr; cp solr.xml zoo.cfg /var/solr/data/); \
-  cp /opt/solr/server/resources/log4j2.xml /var/solr/log4j2.xml; \
-  find /var/solr -type d -print0 | xargs -0 chmod 0770; \
-  find /var/solr -type f -print0 | xargs -0 chmod 0660; \
+  mkdir -p -m0770 /var/solr; \
   sed -i -e "s/\"\$(whoami)\" == \"root\"/\$(id -u) == 0/" /opt/solr/bin/solr; \
   sed -i -e 's/lsof -PniTCP:/lsof -t -PniTCP:/' /opt/solr/bin/solr; \
   chown -R "0:0" /opt/solr-$SOLR_VERSION /docker-entrypoint-initdb.d /opt/docker-solr; \
--- a/8.0/scripts/solr-demo
+++ b/8.0/scripts/solr-demo
@@ -8,6 +8,9 @@ if [[ "${VERBOSE:-}" == "yes" ]]; then
     set -x
 fi
 
+# init script for handling an empty /var/solr
+/opt/docker-solr/scripts/init-var-solr
+
 . /opt/docker-solr/scripts/run-initdb
 
 CORE=demo
--- a/8.0/slim/Dockerfile
+++ b/8.0/slim/Dockerfile
@@ -101,11 +101,7 @@ RUN set -ex; \
   mv /opt/solr/bin/solr.in.cmd /opt/solr/bin/solr.in.cmd.orig; \
   chown root:0 /etc/default/solr.in.sh; \
   chmod 0664 /etc/default/solr.in.sh; \
-  mkdir -p /var/solr/data /var/solr/logs; \
-  (cd /opt/solr/server/solr; cp solr.xml zoo.cfg /var/solr/data/); \
-  cp /opt/solr/server/resources/log4j2.xml /var/solr/log4j2.xml; \
-  find /var/solr -type d -print0 | xargs -0 chmod 0770; \
-  find /var/solr -type f -print0 | xargs -0 chmod 0660; \
+  mkdir -p -m0770 /var/solr; \
   sed -i -e "s/\"\$(whoami)\" == \"root\"/\$(id -u) == 0/" /opt/solr/bin/solr; \
   sed -i -e 's/lsof -PniTCP:/lsof -t -PniTCP:/' /opt/solr/bin/solr; \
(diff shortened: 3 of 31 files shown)

### CODE REVIEW

**Review by makuk66 — APPROVED** [review]
I think that's OK. This will break folks with custom scripts that rely on the current behaviour (and don't run init_var_solr).

Simplifying the printed lines sounds fine too. They were separate mainly to help pinpoint failures, but if you're starting with a blank directory there shouldn't be any, and you can always run with VERBOSE.

**Comment by dsmiley** [comment]
Thanks for your input.  I think this could be released in 8.7 as I think it's minor enough... but I'm fine waiting till 9.0 as well which is perhaps 2-4 months out.

**Comment by nikhildhawan** [comment]
Not sure if I should create a new issue for this but this sounds kind of related so checking here first.

Can you please comment if init_var_solr default location can be the parent directory of $SOLR_HOME rather than fixed /var/solr?

So instead of 
`DIR=${1:-/var/solr}`
It can be
`DIR=${1:-$(dirname "$SOLR_HOME")}`

I am using Azure disk for persistence and want to use precreate-core.
But both precreate-core script and init_var_solr use default /var/solr/data location rather than defined by SOLR_HOME.

Is SOLR_HOME used for completely different purpose?

**Comment by dsmiley** [comment]
Do I get this right?:  So with your proposed change... for most people, there would be no practical difference.  But now it's more feasible to set SOLR_HOME?

Can't you use custom scripts in `/docker-entrypoint-initdb.d` to do whatever pre-creation you want at the standard location?

**Comment by nikhildhawan** [comment]
I have a minimal solr-init.sh file in ` /docker-entrypoint-initdb.d`
```
set -e
echo "Trying to verify adapt core exists before Solr starts" 
precreate-core adapt 
```
but I had to use the default core location for it to work.

precreate-core has this path hard-coded (at least in 8.0-8.6)
`coresdir=/var/solr/data` 
which prevents me from changing the SOLR_HOME to an arbitrary value and using a single command to initialize a blank state.

if both of these files use SOLR_HOME then it shouldn't affect existing users but can make initialization easier by changing SOLR_HOME variable.

I can submit a PR request if it makes sense and if you can guide me on which all versions I should update.

nikhildhawan:
I created https://github.com/docker-solr/docker-solr/pull/357 to show what I am thinking.

makuk66:
Nik: what does Azure have to do with what path is used inside the container?

nikhildhawan:
Hey Martijn.
Not really related to Azure specifically, only for my case I was attaching Azure disk at a place as defined in SOLR_HOME.
I was trying a custom location for it.
During initialization, it was running precreate-core commands successfully but would be blank when starts.
Later I found out the cores were getting created in default location rather than solr_home as per my incorrect assumptions.

The proposed changes will just make it easier to initialize any custom location based on the value of Solr_home.

makuk66:
> I was attaching Azure disk at a place as defined in SOLR_HOME

But why? If there's no good reason, then I don't see why we should make a change in behaviour, which complicates things and can potentially break existing configurations that happen to differ from yours, so would need various scenario testing etc.

nikhildhawan:
It is not about Azure disk, sorry if I made it look that way.
It is about being able to initialize and use any place for the home directory of solr as defined by the Solr_home variable.

If we allow solr_home to be any custom location, but init_var_solr and precreate-core are using the default /var/solr location to create the core then it can cause a conflict.

My simple scenario was a custom Solr_home location and precreate-solr script successfully creating core but at a different(default) location.
I hope I am more clearer now, what do you think @makuk66.

makuk66:
You keep not telling me _why_ you want to do this though.

So let me be more direct: Do not try to set SOLR_HOME to something else. It does not work and the [README](https://github.com/docker-solr/docker-solr/blob/master/README.md#updating-from-docker-solr5-7-to-8) says so:

> The custom SOLR_HOME can no longer be used, because various scripts depend on the new locations

There is no good reason why you'd need to do it, ie there is no use-case it supports that the current mechanisms do not. Using it outside Docker may make sense, but in docker-solr it doesn't. If you want to store things beyond the lifetime of the container, mount a volume on `/var/solr` and you're done. If you want separate data and logs, provide `/var/solr/data` and `/var/solr/logs` and you're done. No need for extra environment variables.

Now, can we change things to make it work even if we do not recommend it? Well, now with David's PR, yes maybe. Your PR is not right though: [here](https://github.com/docker-solr/docker-solr/pull/357/files#diff-f75213b453b244b7973839c4e73fab3f69f41cee9404ef37655d69fee9e671afR17) you're setting SOLR_HOME to `/var/solr` which is wrong; and that file would still need to prepare the log directory in the default place. You made the change in one of the output files instead of the templates and all the output files. And I don't see any test cases that verify that the desired result is achieved. And if that change is made, its use should be explained in the README (and the existing warning removed). I see there are also some mentions of SOLR_HOME in Docker-FAQ.md, so that probably needs verifying and updating too. So, there's a bunch of more work that would need to be done. I'll leave it up to David to decide if that work is worthwhile, and, if so, if that should be part of this PR or a future one.

dsmiley:
I'll just do this in 9.0: https://issues.apache.org/jira/browse/SOLR-15001

nikhildhawan:
Thanks, Martijn.
I didn't pay attention to the mention of SOLR_HOME in README.md. 
It is fine if it is recommended not to change it, I am now using the default location and mounting to /var/solr and it works fine.

My #357 was a draft to describe what I intended to achieve hence wasn't complete and I will close it.

PR #357, opened 2020-10-30
use solr_home for initialisation scripts

nikhildhawan:
Few scripts are using hard-coded locations during initializations that can be changed to SOLR_HOME to allow more flexibility.

### DIFF
--- a/8.0/scripts/init-var-solr
+++ b/8.0/scripts/init-var-solr
@@ -14,7 +14,10 @@ if [[ -n "${NO_INIT_VAR_SOLR:-}" ]]; then
     exit 0
 fi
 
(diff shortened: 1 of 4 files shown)


nikhildhawan:
There are many ways this can be modified for default values and checking directory existence etc.

I am using default location if solr_home is not set in init.
Also setting solr_home if that doesn't exist during precreate, although I am not sure if that can be the case.

nikhildhawan:
Currently, changing SOLR_HOME is not supported as per [README](https://github.com/docker-solr/docker-solr/blob/master/README.md#updating-from-docker-solr5-7-to-8).
Closing this.

PR #359, opened 2020-11-06
Solr 8.7

dsmiley:
https://cwiki.apache.org/confluence/display/SOLR/ReleaseNote87
I followed the "update.md" script, nearly without issue.  My docker image space filled up which caused confusing errors, but once I fixed that, everything built and then tested okay.

### DIFF
--- a/TAGS
+++ b/TAGS
@@ -1,3 +1,5 @@
+8.7:8.7.0:8.7 8 latest
+8.7/slim:8.7.0-slim:8.7-slim 8-slim latest-slim
 8.6:8.6.3:8.6
 8.6/slim:8.6.3-slim:8.6-slim
 8.5:8.5.2:8.5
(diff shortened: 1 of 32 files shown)


TAGS
```diff
@@ -18,3 +18,5 @@
 6.6/slim:6.6.6-slim:6.6-slim 6-slim
 5.5:5.5.5:5.5 5
 5.5/slim:5.5.5-slim:5.5-slim 5-slim
+8.7:8.7.0:8.7 8 latest
+8.7/slim:8.7.0-slim:8.7-slim 8-slim latest-slim
```
makuk66, line 22: These should be sorted at the top; the file appears to be in descending version order. Did the scripts not do that?

makuk66 (changes requested):

TAGS
dsmiley: I simply ran the scripts, following update.md instructions.

makuk66 (approved):

TAGS
dsmiley: I moved it manually.  Can you approve please?

dsmiley:
Here's the PR to official-images: https://github.com/docker-library/official-images/pull/9045

Issue #352, opened 2020-10-09
tools/serve_local.py + tools/build_all.sh instructions assume mac/windows docker features


hossman:
update.md currently suggests...

> To speed up the build by re-using all the Solr tarballs you have locally in ./downloads already, ....
>
> 
> ```
> tools/serve_local.py &
> SOLR_DOWNLOAD_SERVER="http://host.docker.internal:8083" tools/build_all.sh
> wget -t 1 http://localhost:8083/quit >/dev/null 2>&1
> ```

...but "host.docker.internal" didn't work for me.  it is evidently a feature specific to the docker desktop for mac & windows; and not a general feature of docker (on linux for example)...

* https://docs.docker.com/docker-for-mac/networking/#use-cases-and-workarounds
> The host has a changing IP address (or none if you have no network access). We recommend that you connect to the special DNS name host.docker.internal which resolves to the internal IP address used by the host. This is for development purpose and will not work in a production environment outside of Docker Desktop for Mac.

* https://docs.docker.com/docker-for-windows/networking/#use-cases-and-workarounds
> The host has a changing IP address (or none if you have no network access). We recommend that you connect to the special DNS name host.docker.internal which resolves to the internal IP address used by the host. This is for development purpose and will not work in a production environment outside of Docker Desktop for Windows.

----

the workaround i used was to specify a `SOLR_DOWNLOAD_SERVER` pointing at the "real" IP of my machine...

```
tools/serve_local.py &
SOLR_DOWNLOAD_SERVER="http://192.168.1.55:8083" tools/build_all.sh
wget -t 1 http://localhost:8083/quit >/dev/null 2>&1
````

... but AFAICT from various docs, the fact that tools/build.sh specifies `--network=host` when running `docker. build` means that "localhost" from inside docker should point to your "real" host. Sooooo... `SOLR_DOWNLOAD_SERVER="http://localhost:8083"` should also "just work" ... ? maybe?

dsmiley:
I really wish `host.docker.internal` worked universally, or that `--network=host` worked universally.  The situation is such that it's one or the other, depending on your platform :-(

Issue #337, opened 2020-08-13
Stop "supporting" more than two major revisions.


dsmiley:
See the discussion here: https://github.com/docker-library/official-images/pull/8530#discussion_r469892206
> If we cut out anything older than 7.7.x, we should add a warning notice somewhere that these versions are stale, unsupported and likely not secure.

Yes, I think we should do this.  It's consistent with Apache Solr.  If say there's some big bad security bug in Solr for 6.whatever then Solr will get a 6.whatever update and we can do that for docker-solr as well, but that's unusual.  If the security bug is in the JDK... well... I'm not sure it's reasonable for projects to keep releasing updates to old versions to swap in new FROM images.  That seems more like a fundamental limitation in Docker itself -- not allowing you to run image XYZ while swapping out one of the layers (and all it's sub-layers) for another image (and it's sub-layers).  I wish you could!  People should be able to use JDK 15 as soon as it is out if they could.... but they can't (with Docker) without building their own images.  I digress :-)

aadrian:
We are using Docker Solr exactly because it has older versions.

dsmiley:
Quoting the discussion at docker-library:
> but some users still need the older versions/tags, so we don't actively delete them unless the image maintainer explicitly requests that we do so.

So 6.x and 5.x would still be available to you to continue to use; they just wouldn't be updated in normal circumstances.

aadrian:
@dsmiley that's not the same thing like updating them with the latest JDKs and OS.

janhoy:
> @dsmiley that's not the same thing like updating them with the latest JDKs and OS.

True, but if you are concerned with security you won’t be running Solr 5.x in the first place :-) You can check out this repo and build your custom 5.x image with whatever version of Linux and JRE you need.

Perhaps it will be a benefit that the older images will have older build dates on docket hub so it becomes obvious that you’re on your own?

aadrian:
> True, but if you are concerned with security you won’t be running Solr 5.x in the first place :-) 

Many companies use in production older versions of Solr because newer ones are not 100% backwards compatible (drop in replacement) and would require development effort.

Another reason is that Solr itself is used by some other ready project (e.g. Nutch) that can't be just updated without side effects, so people are forced to keep using the older Solr versions.

janhoy:
[...]

PR #361, opened 2020-11-24
use an isolated docker network for server_local.py

hossman:
Having learned a bit more about docker networking since #352 i realized a simple way to solve the issue of running `serve_local.py` in a cross platform way would be to run it in it's own docker container, on an isolated docker network (added bonus: doesn't care if the port is already in use on the docker host)

I implemented a 'wrapper' script to handle spinning up the network & httpd docker container, and shut them down automatically when the build command is finished.

Examples:
```
./tools/serve_local.sh tools/build_all.sh

./tools/serve_local.sh tools/build.sh 8.7
```

### DIFF
--- a/tools/build.sh
+++ b/tools/build.sh
@@ -57,7 +57,7 @@ fi
 if [ "${NOCACHE:-no}" == 'yes' ]; then
   nocache_arg="--no-cache"
 fi
-cmd="docker build --network=host --pull --rm=true ${build_arg:-} ${nocache_arg:-} --tag "$IMAGE_NAME:$full_tag" ."
+cmd="docker build --network=${DOCKER_NETWORK:-host} --pull --rm=true ${build_arg:-} ${nocache_arg:-} --tag "$IMAGE_NAME:$full_tag" ."
 echo "running: $cmd"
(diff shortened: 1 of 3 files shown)


dsmiley:
Looks cool.  It's a bit more advanced relative to what I touch... but if it works, it works.
This script will have a short life-span, if committed -- like only for 8.8.  In 9.0, docker will be published in lucene-solr/solr/docker.  And there's no equivalent need for this over there since the build is computed directly from the built artifacts -- no need to download the tgz.

janhoy:
Cool trick @hossman 

@dsmiley you're right that this repo will likely be rip after 9.0. Depending perhaps on what we do with #337 - i.e. if we want to re-publish 8.x with security-patched Linux and Java versions, or just let them "rot". Also, perhaps the "official" solr repo on docker hub should just forward to the apache/solr one which could be generated by Jenkins.

I guess we should be fine with the latter, as even after 9.0 release we'll likely release 8.8.x bugfix releases, and they will then get an updated Linux/JVM version at the time of release, so users will have a way to run 8.x on semi-current platforms.

PR #364, opened 2021-02-02
Update Readme for solrdata permissions

ninjaaron:
I added a command the to show that the solrdata directory on the host must have the proper permissions to be accessible by the solr user inside the container. In this case, I have done so by changing the user on the host to the same UID as the solr user in the container.


makuk66:
That's not necessary on Mac; you could append a `#` comment to point that out. Not sure about Windows off-hand.

dsmiley:
Thanks!

PR #371, opened 2021-03-05
run `apt-get -y upgrade` on build.

ninjaaron:
Simply running `apt-get update` does not not apply the latest security updates from the Debian---it only refreshes the package index. I would like to build my images locally specifically so I can run on an up-to-date system, but for that, `apt-get upgrade` must also run at build time.

### DIFF
--- a/Dockerfile.template
+++ b/Dockerfile.template
@@ -16,6 +16,7 @@ ARG SOLR_DOWNLOAD_SERVER
 
 RUN set -ex; \
   apt-get update; \
+  apt-get -y upgrade; \
   apt-get -y install acl dirmngr gpg lsof procps wget netcat gosu tini; \
   rm -rf /var/lib/apt/lists/*; \
   cd /usr/local/bin; wget -nv https://github.com/apangin/jattach/releases/download/v1.5/jattach; chmod 755 jattach; \


ninjaaron:
After [further reading](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run) it seems that this is not considered a best practice. I'm not entirely sure what an unprivileged container is, but this does not thrill me greatly.

makuk66:
I found [this example](https://bugs.launchpad.net/ubuntu/+source/libseccomp/+bug/1890332) where someone ran into an issue without using `--privileged`, so perhaps it refers to that sort of thing. I also saw some comments [1](https://github.com/moby/moby/issues/1724#issuecomment-29863741), [2](https://github.com/moby/moby/issues/1724#issuecomment-36466798) where the complaint is that it can be slow and lead to larger-than-required image sizes. And [this comment from 2014](https://web.archive.org/web/20200415055840/http://crosbymichael.com/dockerfile-best-practices-take-2.html) that points out it can lead to image inconsistencies, and specifically points out failing "modify init or make device changes". [here](https://github.com/tianon/docker-brew-ubuntu-core/issues/80) is example of that.

> I would like to build my images locally specifically so I can run on an up-to-date system, but for that, apt-get upgrade must also run at build time.

You could consider building your entire image [FROM scratch](https://docs.docker.com/develop/develop-images/baseimages/#create-a-simple-parent-image-using-scratch), do your own [ubuntu](https://github.com/tianon/docker-brew-ubuntu-core/tree/dist-amd64) and java and solr; that way you're in full control; but obviously that comes with a considerable cost and maintenance responsibility; and how long would you keep that up for?

ninjaaron:
Thanks for the ping back. I'm definitely not building a container from scratch. I just wish there was an easy way to ensure my images are at the latest patch level. Oh well.

Issue #381, opened 2021-06-25
Update generate-stackbrew-library.sh to remove arm64 support for 5.x and 6.x


gadilashashank:
With reference to the discussion [here](https://github.com/docker-library/official-images/pull/10403#discussion_r658957077) support for arm64 was removed in Solr v5.x and v6.x however running the script according to the instructions at [https://github.com/docker-solr/docker-solr/blob/master/update.md#update-the-official-images-repository](https://github.com/docker-solr/docker-solr/blob/master/update.md#update-the-official-images-repository) doesn't reflect this decision.

PR #388, opened 2021-09-30
Special case older solr versions to use hard coded architectures

hossman:
Per comments from smiley on the solr dev list...

> Regarding the arm64v8 tags, those should be removed manually, or fix the
root cause that probably lies in that generate-stackbrew-library.sh
script.  See the PR comments about that here:
https://github.com/docker-library/official-images/pull/10403

### DIFF
--- a/generate-stackbrew-library.sh
+++ b/generate-stackbrew-library.sh
@@ -114,6 +114,11 @@ for version in "${versions[@]}"; do
 		variantParent="$(awk 'toupper($1) == "FROM" { print $2 }' "$dir/Dockerfile")"
 		variantArches="${parentRepoToArches[$variantParent]}"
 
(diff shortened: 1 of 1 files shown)


hossman:
I guess this would fix #381 ?

generate-stackbrew-library.sh
```diff
@@ -114,6 +114,11 @@ for version in "${versions[@]}"; do
 		variantParent="$(awk 'toupper($1) == "FROM" { print $2 }' "$dir/Dockerfile")"
 		variantArches="${parentRepoToArches[$variantParent]}"
 
+    # we don't support the full range of possible archs for these legacy versions...
```
dsmiley, line 117: Looking at this on GitHub suggest the indentation is totally off.  Otherwise LGTM

PR #394, opened 2021-11-30
Replaced wrong URL

Lamera:
The old URL doesn't work anymore.


HoustonPutman:
Since this repo will end with the 8.11.x releases, maybe we use a link to lucene-solr/branch_8_11 instead.

dsmiley (approved):

dsmiley:
Nonetheless I think it's adequate; I don't see us messing with our example data.

PR #396, opened 2021-12-11
Disable log4j message lookup in SOLR_OPTS (CVE-2021-44228)

csandanov:
To mitigate CVE-2021-44228 as instructed at https://solr.apache.org/security.html#apache-solr-affected-by-apache-log4j-cve-2021-44228.

### DIFF
--- a/5.5/scripts/start-local-solr
+++ b/5.5/scripts/start-local-solr
@@ -13,7 +13,7 @@ MY_SOLR_INCLUDE="${SOLR_INCLUDE:-/opt/solr/bin/solr.in.sh}"
 eval "$( set +e +u +o pipefail; . "$MY_SOLR_INCLUDE"; set -o posix; set | grep -E '^SOLR_(LOGS_DIR|OPTS)=' | sed 's/^SOLR_/MY_SOLR_/')"
 
 MY_SOLR_LOGS_DIR="${MY_SOLR_LOGS_DIR:-/opt/solr/server/logs}"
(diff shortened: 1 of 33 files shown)


dsmiley:
Thank you for contributing this!  I'll run tests locally and merge when they pass.

uschindler:
Older versions of Solr use log4j version 1, which is not vulnerable. The system property is not needed there, but does not hurt.

See https://solr.apache.org/security.html#apache-solr-affected-by-apache-log4j-cve-2021-44228 for version details.

dsmiley:
True, but the release automation isn't set up to have peculiarities per-release.  We only have >= 8 and < 8.  The scenario here is rare and in this case harmless so I think it's fine to just release older images as well.

dsmiley:
I should have done my manual test before merging.  The change here doesn't really work, or rather it only helps a scenario that doesn't matter.  It addresses the auxiliary "solr-create" and "solr-demo" scripts since only those call "start-local-solr" which is the script you updated.  The important script is "solr-fg" which starts solr in the foreground.  I'll commit this directly.

dsmiley:
The proper fix: https://github.com/docker-solr/docker-solr/commit/d9aceb632fcad5d5e7ab42b94dd25a008f5b9112
The official-images PR: https://github.com/docker-library/official-images/pull/11485

uschindler:
Did you check with my howto on the solr-security mailing list? (netcat and the emojis?)

dsmiley:
Yes, and thanks for sharing that how-to.  I've used that for other Solr installations to show the issue as well.

News: The docker images are updated on hub.docker.com!

Issue #398, opened 2021-12-15
CVE-2021-44228 Remediation in PR #396 Insufficient according to CVE-2021-45046


pwhite2:
According to CVE-2021-45046 (released today), "previous mitigations involving configuration such as to set the system property log4j2.noFormatMsgLookup to true do NOT mitigate this specific vulnerability". It sounds like you'll need to remove the JndiLookup class from the classpath (example: zip -q -d log4j-core-*.jar org/apache/logging/log4j/core/lookup/JndiLookup.class) or upgrade to log4j 2.16.0 to properly remediate the issue.

plumdog:
From https://github.com/apache/solr/pull/454#issuecomment-994117444, it looks like Solr is not affected and that the original `log4j2.noFormatMsgLookup=true` mitigation is sufficient.

Though, I guess the Docker images could apply the `zip -q -d log4j-core-*.jar org/apache/logging/log4j/core/lookup/JndiLookup.class` mitigation. If Solr isn't using it, this will not break anything and will apply the best mitigation currently known.

pwhite2:
Thanks @plumdog , that would be great - I'm sure that would go a long way towards easing customer concerns.

bmaehr:
An official SOLR 8.11.1 is out

waja:
https://solr.apache.org/docs/8_11_1/changes/Changes.html#v8.11.1.bug_fixes has all the changes ... the relevant part:

> Update Log4J to 2.16

https://issues.apache.org/jira/browse/SOLR-15843

plumdog:
However, as I understand it, there is no intention to release patched earlier versions, eg, no 8.9.1.

dsmiley:
Looks like I commented on the wrong issue yesterday so I will re-do it here for the relevant sentences:

> I'm on top of what's being reported out there, and I don't see a need to post such an issue for CVE-2021-45046 (leading to Log4j 2.16.0) as Solr isn't vulnerable to that. In summary, user input isn't in MDC thus Solr isn't vulnerable. 
> Regardless, the Solr project will upgrade Log4j in 8.11.1 because vulnerability scanners simply can't possibly know what a project is actually vulnerable to based on how it uses the software in question.

> Solr's [security bulletin on this](https://solr.apache.org/security.html#apache-solr-affected-by-apache-log4j-cve-2021-44228) has been updated a lot lately; definitely check it out.

bmaehr:
[...]

Issue #399, opened 2021-12-15
Vulnerable SOLR-Images


nubam:
We still don't see any updates to docker image versions vulnerable to critical CVE-2021-44228 (log4shell).

In our case we're using v7.5.0:
https://hub.docker.com/layers/solr/library/solr/7.5.0/images/sha256-e3db40fa85e7115d2d1d3eb06f7555b6132e33bd3b6e91b17c0a1690122a7acc?context=explore

When will these versions be patched and updated in your docker repository?

baukevdw:
[7.7.3](https://hub.docker.com/layers/solr/library/solr/7.7.3/images/sha256-ac56c3f14054bea4f43ae6e6ed848544e14e178ef436bf5140227496e90fd34e?context=explore) is updated. I don't think they are gonna patch all those old versions. But you can still build the image yourself.

dkd-kaehm:
Due of #282, they will not be updated.
IMHO the images with tags 7.5.x and 7.6.x should be removed, to prevent the folks to start them at all.

adenau:
How about updating 8.11.0 to 8.11.1 ? Is there any time frame?

3XC1T3D:
Or 8.9.0 with updated log4j libs due to new findings of the log4j vulnerability

olstjos:
yes @dsmiley  please cut an 8.11.1 release, I just pulled 8.11 and noticed it still has the vulnerable log4j-api-2.13.2.jar and the log4j-core-2.13.2.jar and other log4j jar files

use this command to test 

```
find / -name '*.jar' | grep log4j
```

HoustonPutman:
We are currently limited by issues with the Apache Infrastructure. The new image will hopefully be released by the end of the day.

Simulant87:
is this the PR to follow? https://github.com/docker-library/official-images/pull/11537 I am not sure which repository is responsible.

HoustonPutman:
Yes that is the PR to follow, unfortunately there are a lot of independent steps in the process that we do not have control over. We had to wait for the apache mirrors to have the artifacts available (took around 16 hours), then now we have to wait for the official docker images team to approve the PR to kick off the official docker image build.

The release process starting in Solr 9.0 should be a little more straightforward, but the delay will always be there when we are registered as an 'Official Image'.

sylus:
Can someone confirm Solr is still ok given the below?

[...]

PR #400, opened 2021-12-16
More brutal CVE-2021-45046 mitigation

plumdog:
See #398.

This is my attempt to apply the fix from CVE-2021-45046, which advises:

> This issue can be mitigated in prior releases (< 2.16.0) by removing the JndiLookup class from the classpath (example: zip -q -d log4j-core-*.jar org/apache/logging/log4j/core/lookup/JndiLookup.class).

I think one of the following must be true:
- Solr makes no use of this code anyway. So this change is both non-breaking (there is no functionality to break) and unnecessary (there are no security holes that this fixes)
- Solr does make use of this code. So this change is both potentially breaking (it may break/change whatever behaviour relies on this code) and necessary (there is a security hole that this fixes, namely CVE-2021-45046)

I think that which of these is true depends on whether the analysis of Solr that the original mitigation for CVE-2021-44228 is sufficient. If it is, then the former is true, otherwise the latter.

My view is that in either case, the Solr docker image is better with this change applied, hence my opening of this PR. I appreciate there may well be concerns about this change (hacking some `jar`s with zip :shrug:, maybe it will break other things for reasons I don't understand, maybe the risk of breaking some legitimate uses of JndiLookup outweigh the perceived risks of the security flaw, maybe Solr releases are inbound to bump log4j to 2.16.0).

Also, I have only tested that with this change, I can run `./tools/build_latest.sh` and the docker build succeeds, and that when I run the image it built, it seems to start Solr and I can see the Solr UI in a browser.

### DIFF
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -16,7 +16,7 @@ ARG SOLR_DOWNLOAD_SERVER
 
 RUN set -ex; \
   apt-get update; \
-  apt-get -y install acl dirmngr gpg lsof procps wget netcat gosu tini; \
+  apt-get -y install acl dirmngr gpg lsof procps wget netcat gosu tini zip; \
   rm -rf /var/lib/apt/lists/*; \
   cd /usr/local/bin; wget -nv https://github.com/apangin/jattach/releases/download/v1.5/jattach; chmod 755 jattach; \
   echo >jattach.sha512 "d8eedbb3e192a8596c08efedff99b9acf1075331e1747107c07cdb1718db2abe259ef168109e46bd4cf80d47d43028ff469f95e6ddcbdda4d7ffa73a20e852f9  jattach"; \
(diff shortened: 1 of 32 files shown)


bmaehr:
An official SOLR 8.11.1 is out

waja:
https://solr.apache.org/docs/8_11_1/changes/Changes.html#v8.11.1.bug_fixes has all the changes ... the relevant part:

> Update Log4J to 2.16

https://issues.apache.org/jira/browse/SOLR-15843

luis-serra-ki:
This is a high priority fix. Merge it guys, please

janhoy:
Apache Solr is not affected by CVE-2021-45046, as explained in https://solr.apache.org/security.html#apache-solr-affected-by-apache-log4j-cve-2021-44228

Further, "If you are using Solr's official docker image, it has already been mitigated in all versions listed as supported on Docker Hub: https://hub.docker.com/_/solr. You may need to re-pull the image."

jordonedavidson:
As of this comment the official Docker Solr image (referenced by @janhoy ) is still at v.8.11.0 which is vulnerable per the Solr docs. Any Idea when this image will be updated?

HoustonPutman:
It will be available by the end of the day hopefully. There are issues with the Apache infra that we are working through currently.

olstjos:
@dsmiley 

https://github.com/docker-solr/docker-solr/issues/399#issuecomment-996039940

waja:
accordingly to https://solr.apache.org/security.html#apache-solr-affected-by-apache-log4j-cve-2021-44228:

> If you are using Solr's official docker image, it has already been mitigated in all versions listed as supported on Docker Hub: https://hub.docker.com/_/solr. You may need to re-pull the image.

olstjos:
Mitigation or not, people responsible for security are nervous and knee jerking.  For example our organization has put in restrictions so that we can no longer deploy images with the vulnerable Log4J library so for us to be able to deploy Solr we will now require 8.10.1

dsmiley:
> As of the time of this comment the official Docker Solr image (referenced by @janhoy ) is still at v.8.11.0 which is vulnerable per the Solr docs

The Solr docs declare it is vulnerable?!  Please share a reference.  Perhaps you mean that the Solr news / security bulletin at some point in time said Solr 8.11 was vulnerable.  It did, and since that time, it has been updated (quoted above) based on mitigations performed in our docker images.  What's cool to see is that we can update older releases at the docker layer of the release without going back to the original release branches (e.g. for 7x).  I know this can complicate the story about knowing if you are vulnerable or not.  

Any way; I'm working on the 8.11.1 image.  I'll have a PR for that in a bit.

3XC1T3D:
Will there be also an log4j Update for the < 8.11.1 Images such as 8.9.0?

dsmiley:
> Will there be also an log4j Update for the < 8.11.1 Images such as 8.9.0?

Probably not.

PR #401, opened 2021-12-16
Solr 8.11.1 upgrade.  Explain Log4J situation.

dsmiley:
I updated the README to have a section about this at the very top.  We'll likely move this down in some months time.


janhoy:
This must be wrong. The PR contains a ton of test log output...

HoustonPutman:
I like the notice about the CVEs. But yeah, delete the test logs.

README.md
```diff
@@ -1,7 +1,17 @@
+# NOTE: Not vulnerable to Log4J 2 "Log4shell"
+
+The Docker images *were* vulnerable to one of a pair of vulnerabilities in Log4J 2 but they are not vulnerable anymore -- you may need to re-pull the image you are using.  For images prior to 8.11.1, Solr is using a popular technique to do this -- setting `log4j2.formatMsgNoLookups`.  The Solr maintainers have deemed this adequate based specifically on how Solr uses logging.  It won't be adequate for all projects that use Log4J.  Scanning software might alert you to the presence of an older Log4J JAR file, however it can't know if your software (Solr) uses the artifacts in a vulnerable way.
```
janhoy, line 3: > The Docker images *were* vulnerable to one of a pair of vulnerabilities in Log4J 2 but they are not vulnerable anymore

A lot of the images on hub are still vulnerable. If someone uses e.g. `solr:7.6` they are still vulnerable. They need to upgrade to the latest minor release, e.g. 5.5, 6.6, 7.7, or the latest bugfix release on 8.x, e.g. 8.10.1, but not e.g. [8.10.0](https://hub.docker.com/layers/solr/library/solr/8.10.0/images/sha256-46fef455acab41b7d555638cfa1e1314cb7eb7fd5667302018e5c0e80743ef86?context=explore).

Btw - I see that docker hub has started scanning images for log4shell, perhaps they look for `log4j-core-2.x.jar`. Don't know if they will ban pulls for vulnerable images or what.

TAGS
```diff
@@ -1,5 +1,5 @@
-8.11:8.11.0:8.11 8 latest
-8.11/slim:8.11.0-slim:8.11-slim 8-slim latest-slim
+8.11:8.11.1:8.11 8 latest
+8.11/slim:8.11.1-slim:8.11-slim 8-slim latest-slim
```
janhoy, line 2: Shall we stop refreshing 5.x, 6.x and 7.x already now? Then they will fall off the "supported" list. The tags will still be available, but not promoted.

janhoy (changes requested):

TAGS
olstjos: I suggest to have this debate after 8.11.1 is released because open ended questions during a security update just delays action.  Even if it's mitigated, it'd be nice to get the vulnerable Log4J out of a build.  Security people all over the place are watching and blocking things that throw red flags even if they're 'mitigated'.

olstjos (approved):

TAGS
janhoy: Yes, we can fix #337  separately, but it is unfortunate to pretend that 5 year old versions are "supported"

README.md
dsmiley: Maybe you are forgetting/unaware, but there is a mitigation here (in docker-solr) for this: https://github.com/docker-solr/docker-solr/commit/d9aceb632fcad5d5e7ab42b94dd25a008f5b9112 applies to all our tags (thus 8.10 etc.).

TAGS
dsmiley: _Right_ now doesn't seem the time to do that.

README.md
dsmiley: I'm concerned I had to explain that.  Was the explanation in the text on the page not sufficient?  Maybe I could incorporate a link to the evidence (my commit)?

dsmiley: I'm concerned I had to explain that.  Was the explanation in the text on the page not sufficient?  Maybe I could incorporate a link to the evidence (my commit)?

dsmiley:
The files I accidentally committed were numerous, so I reset to the previous branch completely and re-updated the images, and I added in the README.md file from before.  Sorry if this kind of reset the review but I didn't make other changes.
BTW I ran all tests locally and they passed.

HoustonPutman (approved):
Looks good to me

dsmiley:
Official-Images PR: https://github.com/docker-library/official-images/pull/11537

8.11/slim/Dockerfile
```diff
@@ -4,9 +4,9 @@ FROM openjdk:11-jre-slim
 LABEL maintainer="The Apache Lucene/Solr Project"
 LABEL repository="https://github.com/docker-solr/docker-solr"
 
-ARG SOLR_VERSION="8.11.0"
-ARG SOLR_SHA512="fa766775a70ee636792149baa20a541fa043a7579d499324072965f0c241600b75afe940d2b6c90cafb4b14b9c5f1c2a7bd0844c7d198ba9a36f09e427262dd2"
-ARG SOLR_KEYS="E6E21FFCDCEA14C95910EA65051A0FAF76BC6507"
+ARG SOLR_VERSION="8.11.1"
+ARG SOLR_SHA512="4893f836aec84b03d7bfe574e59e305c03b5ede4a48020034fbe81440b8feee79e55fd9ead230e5b89b3f25124e9b56c1ddc4bb5c7f631cf4e846b9cab5f9a45"
+ARG SOLR_KEYS="2CECBFBA181601547B654B9FFA81AC8A490F538E"
```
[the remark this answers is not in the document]
janhoy: Yes, it's my key, and I see that Dockerfile just search keyservers for it (and then it will fetch the entire key), so validation passes.

README.md
janhoy: > Maybe you are forgetting/unaware, but there is a mitigation here (in docker-solr) for this

Nope, this commit does NOT patch v 7.0, 7.1, 7.2, 7.3, 7.4, 7.5, 7.6. As I say, only the latest minor version of 5.x, 6.x and 7.x are patched, so it is dangerous to write "but they are not vulnerable anymore" as a blanket statement.

dsmiley: Ooooh, ok I see your point!  Sorry.  We should communicate the range of Docker images that are vulnerable (which does _not_ include 7.0, 7.1, 7.2, 7.3, that you listed).  It's 7.4 to < 7.7.3.  The recommendation for such users is to upgrade to 7.7.3 (specifying Docker tag 7 or 7.7 is fine) or set that system property.

I will update the README now.

dsmiley: #402

PR #402, opened 2021-12-17
Readme clarification; what is vulnerable

dsmiley:


README.md
```diff
@@ -1,6 +1,6 @@
...
+The Docker images *were* vulnerable to one of a pair of vulnerabilities in Log4J 2 but they are not vulnerable anymore -- you may need to re-pull the image you are using.  Unmaintained images (not listed in TAGS, and thus not listed on [Docker Hub](https://hub.docker.com/_/solr?tab=tags), e.g. versions between 7.4 inclusive and 7.7.3 exclusive and some later ones) *are* vulnerable.  For supported images prior to 8.11.1, Solr is using a popular technique to mitigate the problem -- setting `log4j2.formatMsgNoLookups`.  The Solr maintainers have deemed this adequate based specifically on how Solr uses logging.  It won't be adequate for all projects that use Log4J.  Scanning software might alert you to the presence of an older Log4J JAR file, however it can't know if your software (Solr) uses the artifacts in a vulnerable way.  If you are in doubt of the mitigation being in place, look for `-Dlog4j2.formatMsgNoLookups` in the Args section of Solr's front admin screen.
```
janhoy, line 3: Suggestion that cannot be misunderstood:

All Solr Docker images *were* vulnerable to one of a pair of vulnerabilities in Log4J 2. But we have patched and re-built all *supported* image tags. So make sure you are pulling only those tags listed as supported on [Docker Hub](https://hub.docker.com/_/solr?tab=tags). You may need to re-pull the image you are using. For supported images prior to.....

janhoy (commented):
I think we should not confuser users with version-ranges etc, when it is easier to say that *only* the supported tags are patched.

janhoy (approved):

PR #403, opened 2021-12-17
Only refresh 7.7 and 8.11 versions

### LINKED ISSUE #337: Stop "supporting" more than two major revisions.

janhoy:
Fixes #337 

This PR removes all tags except `8.11` and `7.7`. These are the two versions that are officially supported and updated with bug-fixes and security-fixes in solr, so it should be the same on docker. After this change, only two tags will be listed on the "supported" list. It will still **work** to select other historic tags, but they are more hidden and the timestamp on those will reflect when the CODE was last updated. Hopefully that will discourage users from picking v7.5, 8.1 or similar.

The PR also updates the list of GPG keyservers to use secure transport and remove the sks-keyservers that are deprecated.

NOTE: update.sh failed for me, so I hand edited `TAGS` file in this PR. I have not run the tests either.


dsmiley:
I feel hesitant to remove the earlier 8.x releases right now but I understand your rationale and I don't feel strong enough to block it.  I'm glad we had these releases over the last week for the sake of our users.  Maybe it was a special circumstance.

janhoy:
Advantage of refreshing 8.0, 8.1, 8.2, 8.3, 8.4, 8.5, 8.6, 8.7, 8.8, 8.9 and 8.10 for every 8.11.x release is that those images get refreshed with latest bugfix java-11 and Linux.

Disadvantage is that these old releases are in practice not supported and patched by the PMC. Solr 8.0.0 is from **march 2019** and is vulnerable to [7 critical/high CVEs](https://nvd.nist.gov/vuln/search/results?adv_search=true&isCpeNameSearch=true&query=cpe%3A2.3%3Aa%3Aapache%3Asolr%3A8.0.0%3A*%3A*%3A*%3A*%3A*%3A*%3A*).

I think we should not advertise (support) other than the latest point release for the two currently supported major versions. But if others have strong arguments FOR promoting these old releases, I'm all ears.

dsmiley (approved):

janhoy:
Would love feedback from @makuk66 , @HoustonPutman and @hossman too on this. Also in context of how we do this from 9.0 on

makuk66:
> I think we should not advertise (support) other than the latest point release for the two currently supported major versions. 

Certainly agreed about the "advertise" part.

It'd be nice if the decision that to rebuild could be orthogonal (pick up new base images, add environmental security mitigations to Dockerfile etc) but given how the builds and docker hub listing works, I guess that's not really possible.

I've tended to err on the side of doing refresh older images; mainly because I've seen folks running old versions in situations that could not be upgraded to new major versions because of config changes or jar incompatibilities, or lack of resources, or organisational barriers to deployment etc. But especially in light of this week's events, perhaps enabling that unresponsive behaviour is ultimately actually a disservice.

> Also in context of how we do this from 9.0 on

If 9 is a big change, which I expect it will be, especially if there are difference decisions about base image selection, then I suggest extending the "two currently supported major versions" to 3, during the first few minor releases of 9 at least.

janhoy:
>If 9 is a big change, which I expect it will be, especially if there are difference decisions about base image selection, then I suggest extending the "two currently supported major versions" to 3, during the first few minor releases of 9 at least.

If a 7.7.4 release ever happens then I don't see a problem extending "support" for 7.7.x for some time. But as long as 7.7 is stuck at 7.7.3 (april 2020) we should stop refreshing it too.

HoustonPutman:
I agree it's unfortunate that we can't keep all images up to date without advertising them.

Maybe we instead just support solr versions released in the last year (or 18 months, whatever we want the timebox to be)? That would give us 8.8.0 -> 8.11.1 currently, giving people a year to upgrade to the next version to keep receiving operating system upgrades, etc. This might be harder to code, but it would at least be more flexible than "The last minor version", and actually have an SLA for people to follow, and a timeline that they need to upgrade after.

janhoy:
> Maybe we instead just support solr versions released in the last year (or 18 months, whatever we want the timebox to be)? That would give us 8.8.0 -> 8.11.1 currently, giving people a year to upgrade to the next version

That could be one way to go. In our [download page ](https://solr.apache.org/downloads.html#about-versions-and-support)we say:

> Apache Solr is under active development with frequent feature releases on the current major version. The previous major version will see occasional critical security- or bug fixes releases. 

This currently sends a signal that if you value stability you should pick 8.11.x, but if you want new features you should stay on latest 9.x minor release. We never promise to bugfix each intermediate minor version, so once 9.1 is released, 9.0.x will **not** be updated. I can see that before docker, users would regularly patch linux and jdk on their servers, but typically stay on a stale Solr version. We have given them that option by refreshing. But I wonder if that mentality belongs to another era now. Solr has more than 100 jar dependencies, and vulnerabilities are bound to occur. So our users may be better off staying more up to date also with the Solr version.

So perhaps auto refresh last 3 minor versions on stable line. But if a serious CVEs occurs, we nullify that and only push the patched versions to hub? Then on a case-by-case basis we could also consider releasing bugfix versions of a previous minor version, to provide users with a more stable bugfix upgrade.

I wonder how many users specify `8` vs `8.11` vs `8.11.1` in their docker tags? Specifying `8.11` would select the latest patch version on redeploy. But solr-operator would need explicit full tags to handle rolling upgrade? Would be cool if operator would poll for new bugfix releases in the background if provided with `8.11` :)

janhoy:
I can wait with this until 9.0 is out, then only 9.0.0 and 8.11.1 will be advertised.

Re. advertising when refreshing - if another log4shell kind of incident happens, we may perhaps have a workflow where we temporarily push a change to a bunch of image versions, to refresh them, but immediately after go back to the two "supported" versions and push those. That way the old tags gets refreshed with whatever mitigations and linux+java-updates there are, but they don't remain advertised...

HoustonPutman:
@janhoy Do we want to move the 8.11 directory to https://github.com/apache/solr-docker? That way we can just make this repository read-only, and all development will be done in the apache/solr-docker repo? We don't use any of the scripting anymore, I believe... We are just manually changing the patch version, SHA and GPG key on future 8.11.x releases.

janhoy:
Why not! If there is a new 8.11.x release the new workflow must be used anyway, so can just as well have the master data in Solr repo. I’ll close this PR.

PR #406, opened 2021-12-27
Mac toolchain updates

acdha:
This has two minor changes to the build process on macOS which I encountered while working on #405:

* Add the `gnu-sed` requirement to update.md to avoid an error round-trip
* Support Homebrew's [default prefix change for ARM](https://docs.brew.sh/FAQ#why-should-i-install-homebrew-in-the-default-location)


janhoy:
Thanks for the nice contrib

PR #426, opened 2022-06-23
Solr 8.11.2 upgrade

colvinco:
Move 8.11.1 to 8.11.2


colvinco:
Sorry I missed https://github.com/docker-solr/docker-solr/blob/master/update.md so there may be some steps I haven't followed, I will check that

PR #427, opened 2022-08-03
SOLR-16323: Change USER to use ID instead of Name

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-16323

This will also be landing in Solr 9.1


dsmiley:
I'm not familiar with the details but can you attest that this is completely backwards compatible?

HoustonPutman:
Closing and continuing in https://github.com/apache/solr-docker/pull/9

Will address your question there @dsmiley

